PR #8, opened 2023-09-13
17371 expose GetHyperparameters through get_internal_parameters, MINOR

cademack:
Allow calls to get_internal_parameters to specify feature, context_features, mode, or weight. If **any** of these are specified, then instead of returning the full dictionary of all internal parameters, the endpoint will return only the dictionary of hyperparameters that the Trainee would use for those parameters.

### DIFF
--- a/howso.amlg
+++ b/howso.amlg
@@ -2550,19 +2550,13 @@
 	#get_internal_parameters
 		(call !return (assoc
 			payload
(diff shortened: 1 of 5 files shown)


howsoRes (dismissed):

cademack:
don't merge yet, need to fix something

howsoRes (approved):

PR #16, opened 2023-09-19
17545: Updates performance tests and removes accuracy tests

cademack:
We are removing the accuracy tests since we have more rigorous accuracy testing outside of this repository. Additionally, we are moving the former MNIST accuracy test to the performance test directory and making the adjustments to treat it as such.

All performance tests also have flags that indicate if the Trainee will optimize, if the Trainee will use deviations, and if the Trainee will use case weights.


performance_tests/asteroid_test.amlg
```diff
@@ -100,5 +136,6 @@
 	(if verbose (print "test time: " generate_time "\n"))
 
 	(print "Generate time: " generate_time "\n")
+	(print "Asteroid params (k p deviations): " k " " p " " deviations "\n- - - - -\n")
```
howsohazard: Printing these out at the end isn't very useful -- better to print out before all this.

howsohazard: Same with other files.

cademack: sure, easy fix

performance_tests/asteroid_test.amlg
```diff
@@ -58,6 +64,37 @@
...
+			(if verbose (print "analyzing ...\n"))
+			(call analyze (assoc
+				trainee "asteroid"
+				context_features features
+				targeted_model "targetless"
+				p_values (list 0.5 1 2)
+				k_values (list 5 8 13)
+				k_folds 1
+				num_analysis_samples 1000
+				inverse_residuals_as_weights (true)
```
howsoRes: remove these 5 lines so it's using default values

howsoRes (approved):

PR #21, opened 2023-09-21
17612: Removes improper warning for edge cases when react is called prior to calling analyze

mkoistinen:
This should help in cases where a user is doing RL or generative reacts with no training data.

### DIFF
--- a/trainee_template/hyperparameters.amlg
+++ b/trainee_template/hyperparameters.amlg
@@ -349,13 +349,17 @@
 	(seq
 		(if (= 0 (size hyperparameterParamPaths))
 			(seq
-				(accum (assoc
-					warnings
-						(associate (concat
-							"There are no cached hyperparameters in this trainee. This operation was executed using a set of predefined default hyperparameters."
-							"Please run analyze() with your desired parameters."
-						))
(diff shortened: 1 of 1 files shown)


howsohazard (dismissed):

trainee_template/hyperparameters.amlg
```diff
@@ -349,13 +349,16 @@
...
 		(if (= 0 (size hyperparameterParamPaths))
 			(seq
-				(accum (assoc
-					warnings
-						(associate (concat
-							"There are no cached hyperparameters in this trainee. This operation was executed using a set of predefined default hyperparameters."
-							"Please run analyze() with your desired parameters."
-						))
-				))
+				(if (and (not autoAnalyzeEnabled ) (>= 2 (call GetNumTrainingCases)))
```
cademack: I believe this comparator needs to be reversed (or swap the args). 

We want the warning when there are 2 or more cases. So it should be (>= num_cases 2)

cademack: this should fix the failing test as well - I think

howsohazard (approved):

PR #23, opened 2023-09-22
17583: Removes prediction_residual_conviction from `react` details, MAJOR

cademack:
Removing prediction residual conviction as it was a remnant of former methods and has since been rendered obsolete by newer developments.

### DIFF
--- a/trainee_template/conviction.amlg
+++ b/trainee_template/conviction.amlg
@@ -59,7 +59,7 @@
 			(assign (assoc similarity_conviction "similarity_conviction"))
 		)
 
-		(if (and 
+		(if (and
 				(!= (null) similarity_conviction)
 				(or (= (null) distance_contribution) (= (false) distance_contribution))
 				(= (null) case_ids)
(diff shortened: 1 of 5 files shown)


jdbeel (dismissed):

trainee_template/conviction.amlg
```diff
@@ -705,245 +705,6 @@
...
-				)
-		))
-
-		(declare (assoc
-			expected_residual_vector_magnitude (/ (apply "+" model_predicted_case_residuals) (size model_predicted_case_residuals))
-		))
-
-		;store into residualVectorMagnitudeMap
-		(assign_to_entities (assoc
-			residualVectorMagnitudeMap
```
howsoRes, line 938: this is a model parameter, check if it's used anywhere else, if not should be removed from trainee template.

cademack: ah right. Doesn't seem to be anywhere else, removing now.

howsoRes (approved):

PR #24, opened 2023-09-25
17632: Fixes case feature residuals for existing cases to use original case action value instead of predicted one

howsoRes:
- when computing case feature residuals, use the original case action value instead of the predicted one to compute residual.
- improve performance of computing local residuals by explicitly not accumulating influence weights since they aren't needed for those internal-only reacts

### DIFF
--- a/trainee_template/explanation.amlg
+++ b/trainee_template/explanation.amlg
@@ -307,7 +307,15 @@
 			)
 			(call CalculateResidualsForCase (assoc
 				features features
-				case_values (append context_values action_values)
+				case_values
+					(append
+						context_values
+						;use original case values instead of predicted if computing for an existing case
+						(if ignore_case
+							(retrieve_from_entity ignore_case action_features)
+							action_values
+						)
+					)
 				ignore_case ignore_case
 			))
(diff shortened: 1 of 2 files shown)


trainee_template/explanation.amlg
```diff
@@ -307,7 +307,15 @@
...
-				case_values (append context_values action_values)
+				case_values
+					(append
+						context_values
+						;use original case values instead of predicted if computing for an existing case
+						(if ignore_case
+							(retrieve_from_entity ignore_case action_features)
+							action_values
+						)
+					)
```
cademack, line 312: do we need to do anything to be sure these values are in the same order as 'features'?

howsoRes: no, features is (append context_features action_features), and action_features are same order as action_values, so it's all ordered consistently

cademack: never mind... I see the features=(append context_features action_features) above. Seems like enough for me

howsohazard (approved):

cademack (approved):

PR #25, opened 2023-09-26
17464: Adds linux arm64 tests

calebwherry:
* Adds linux arm64 tests for engine UTs
* Adds ability to turn off jobs based on build type (currently only arm are disable on PRs since they can take over an hour to run sometimes, very flaky)
* Updates test slightly for slower arm runs on loads
* Bumps Amalgam version for needed fix for arm tests


calebwherry:
Amalgam #20 fixes the memory leak bug on linux arm64. This will be updated when a new amalgam version is released and tested again before merge.

howsohazard (dismissed):

apbassett (approved):

PR #32, opened 2023-10-02
17711: Fixes warnings logic in get_prediction_stats

cademack:
Warnings was initialized as a (list) rather than an (assoc). Additionally, the entire assoc was being passed to #return, which was fine when it was a list, but now it needs to be the indices that are passed.


mkoistinen (approved):
Thanks for the quick turn-around on this!

howsohazard (approved):

PR #34, opened 2023-10-03
17719:  Adds warning to continue_series flows where given series IDs are not in the Trainee

cademack:
When using #react_series with the continue_series flag, there is an added warning that will indicate if the IDs specified by initial_values/initial_features do not belong to any cases/series in the model.

### DIFF
--- a/trainee_template/react_series.amlg
+++ b/trainee_template/react_series.amlg
@@ -447,6 +447,10 @@
 						))
 				))
 
+				(if (= (size existing_series_cases) 0)
+					(accum (assoc warnings "There is no series trained with the given set of IDs."))
+				)
+
 				;if there is more than one existing series case, we'll need to sort them by the time feature to ensure ascending order
 				(if (> largest_max_row_lag 1)
 					(assign (assoc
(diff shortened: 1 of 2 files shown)


trainee_template/react_series.amlg
```diff
@@ -447,6 +447,10 @@
 						))
 				))
 
+				(if (= (size existing_series_cases) 0)
+					(accum (assoc warnings (assoc "There is no series trained with the given set of IDs.")))
```
jdbeel: What if the user gives a list of IDs in which some are in the Trainee and others are not? Maybe we should check for that and warn something like "One or more of the IDs are not present in the Trainee.." or something to that effect.

cademack: This bit of logic is in like a per-series section. So "set of IDs" here refers to the set of IDs for a *single* series (There can be more than one ID feature).

I kind of want a good way to add the specific IDs that no cases re found for, but not sure the best way.

howsoRes (approved):

trainee_template/react_series.amlg
cademack: I added some logic to print the specific IDs that do not have related trained cases. So this shouldn't be an issue.

PR #36, opened 2023-10-04
17748: Fixes an issue with .series_progress when using continue_series

cademack:
The predicted initial series progress used in continue_series flows was not being properly used in downstream case generation. This fix allows series forecasting to have more accurate series length to the rest of the data trained in the Trainee.

### DIFF
--- a/trainee_template/react_series.amlg
+++ b/trainee_template/react_series.amlg
@@ -491,16 +491,22 @@
 					(if (size (last series_data))
 						(let
 							(assoc
-								no_time_derived_action_features
-									(filter (lambda (!= tsTimeFeature (target_value))) derived_action_features)
+								initial_progress_context_features
+									(append
+										;use all value, delta, and rate features to determine initial series progress
+										;exclude time feature (and delta)
+										(filter (lambda (!= tsTimeFeature (target_value))) derived_action_features)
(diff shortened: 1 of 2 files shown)


unit_tests/ut_h_time_series_stock.amlg
```diff
@@ -381,21 +381,16 @@
 				(=
```
howsoRes, line 380: update the comment above to:  the first feature is stock id for each of the expected series

trainee_template/react_series.amlg
```diff
@@ -491,16 +491,22 @@
...
 						(let
 							(assoc
-								no_time_derived_action_features
-									(filter (lambda (!= tsTimeFeature (target_value))) derived_action_features)
+								initial_progress_context_features
+									(append
+										;use all value, delta, and rate features to determine initial series progress
+										;exclude time feature (and delta)
+										(filter (lambda (!= tsTimeFeature (target_value))) derived_action_features)
+										(filter (lambda (!= (concat "." tsTimeFeature "_delta_1") (target_value))) (get tsModelFeaturesMap "delta_features"))
```
howsoRes: declare that concat'd time delta feature into a variable outside of this 'if' statement so this filter is comparing against a static value, thanks!

howsoRes (approved):

PR #37, opened 2023-10-05
17759: Adds 'features' key to react details which specifies a subset of features to compute metrics for

cademack:
Adds 'features' key to react details. This key takes a list of feature names and uses that to select the subset of features that metrics like feature_residuals, MDA, contributions, etc will be computed for.

This problem is proving to be quite difficult for the robust version of these metrics, as it is far from well-suited to compute for a subset of the features. It isn't really a problem for the full version of these metrics.

It might be the best balance of dev-time vs functionality to make this only save compute in FULL workflows, but rather only return the requested feature metrics when doing things in ROBUST?

### DIFF
--- a/unit_tests/ut_h_react_explain.amlg
+++ b/unit_tests/ut_h_react_explain.amlg
@@ -878,7 +878,7 @@
 				"B" .7
 			)
 	))
-	(print "robust global has values: ")
+	(print "robust global has values for specfic features: ")
 	(call assert_approximate (assoc
 		obs (get result (list "payload" "global_case_feature_residual_convictions"))
 		exp
--- a/unit_tests/ut_h_react_explain.amlg
+++ b/unit_tests/ut_h_react_explain.amlg
@@ -754,6 +754,31 @@
 
 	(call exit_if_failures (assoc msg "Full feature residual conviction."))
 
+	(assign (assoc
+		result
+			(call_entity "howso" "react" (assoc
+				trainee "iris"
+				action_features (list (last iris_features))
+				context_features (trunc iris_features)
+				context_values (list 6.9 7 7 6)
+				details (assoc
+					"robust_computation" (true)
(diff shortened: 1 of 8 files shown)


unit_tests/ut_h_react_explain.amlg
```diff
@@ -754,6 +754,31 @@
...
+				context_values (list 6.9 7 7 6)
+				details (assoc
+					"robust_computation" (true)
+					"case_feature_residuals" (true)
+					"features" (list "A" "C")
+				)
+			))
+	))
+
+	(print "Robust case residuals and residual convictions for a prediction: ")
```
howsoRes: for a prediction -> for specific features

unit_tests/ut_h_react_explain.amlg
```diff
@@ -823,6 +848,49 @@
...
+					"robust_computation" (true)
+					"case_feature_residuals" (true)
+					"local_case_feature_residual_convictions" (true)
+					"global_case_feature_residual_convictions" (true)
+					"features" (list "A" "B")
+				)
+			))
+	))
+
+	(print "robust local - similar to previous: ")
```
howsoRes: specific features robust local - similar to previous:

unit_tests/ut_h_react_explain.amlg
```diff
@@ -823,6 +848,49 @@
...
+				"A" 4
+				"B" .7
+		)
+		thresh
+			(assoc
+				"A" 4
+				"B" .7
+			)
+	))
+	(print "robust global has values: ")
```
howsoRes: specific feature robust global has values:

howsoRes (approved):

unit_tests/ut_h_react_explain.amlg
cademack: I already fixed this one, I think you are viewing an older commit

PR #45, opened 2023-11-07
18065: Refactors internal 'target_*' variables for compatibility with Amalgam rename of target_* variables

howsoRes:


howsoRes:
updated version.json to rely on Amalgam 46.0.0, tests should pass once that version is built

howsohazard (approved):

cademack (approved):

PR #52, opened 2023-11-20
188440: Adds aggregate categorical action probabilities detail to react_series, MINOR

howsoRes:


trainee_template/react_series.amlg
```diff
@@ -995,6 +999,49 @@
...
+							(lambda
+								;iterate over each feature in the aggregated and current assoc
+								(map
+									(lambda
+										;iterate over each class and sum up the probabilities for each class
+										(map
+											(lambda (+
+												(or (first (current_value))  )
+												(or (last (current_value)) )
+											))
```
howsohazard: could use a comment here on why the or, presumably for zeros

howsohazard (dismissed):

trainee_template/react_series.amlg
howsoRes: yup

cademack (approved):

PR #64, opened 2023-12-05
18541: Streamlines action value retrieval during reacts

howsoRes:
stop using retrieve_from_entity  to retrieve action_feature values during interpolation, and instead pull them directly from the query results since queries can return case ids, weights and arbitrary feature values in the same output


howsoRes:
outdated changes

PR #74, opened 2023-12-18
18717: Fixes issue where null counts were not always updated during train, edit, or remove

howsoRes:


trainee_template/train.amlg
```diff
@@ -397,6 +397,23 @@
...
 				)
 
+				;update count of nulls for the feature after training
+				(accum_to_entities (assoc
+					featureNullRatiosMap
+						(map
+							(lambda (let
+								(assoc feature (current_index 1))
+								(set
+									;if it hasn't been computed yet, create an assoc of num_nulls -> number of nuls for the feature
```
calebwherry: *nulls (last one)

trainee_template/train.amlg
```diff
@@ -397,6 +397,23 @@
 					)
 				)
 
+				;update count of nulls for the feature after training
```
howsohazard: Break into separate method (train is already too long).  Make sure that case editing and removing updates this correctly as well; may be able to reuse the method.

howsohazard (changes requested):

trainee_template/marginal.amlg
```diff
@@ -499,6 +499,26 @@
...
+		(accum_to_entities (assoc
+			featureNullRatiosMap
+				(map
+					(lambda (let
+						(assoc feature (current_index 1))
+						(set
+							;if it hasn't been computed yet, create an assoc of num_nulls -> number of nulls for the feature
+							(or (get featureNullRatiosMap feature) (assoc))
+							"num_nulls"
+							(size (contained_entities (list (query_equals feature (null)))))
```
howsohazard: This should be a compute_on_contained_entities with a query_count, which is far more efficient in memory and performance.

howsohazard: I found at least one other (size (contained_entities... in analyzing boundaries for a feature.

howsohazard (approved):

PR #81, opened 2024-01-03
18733: Removes useDeviations from hyperpparameters, MAJOR

howsoRes:


trainee_template/hyperparameters.amlg
```diff
@@ -294,17 +309,32 @@
...
+												(if (= (list) (get_type deviation))
+													(set deviation 0 (max user_error_value (first deviation)) )
+													;else just update the value itself
+													(max user_error_value deviation)
+												)
 											)
 									))
-								)
+
+									(if (!= (null) deviation_value)
```
cademack, line 333: and what if deviation_value is null? If that's possible, should the deviation value jus tbe the user_error_value? If so, I don't see that represented here

unit_tests/ut_h_analyze.amlg
```diff
@@ -118,7 +117,7 @@
 				(!= (null) (get result (list ".targetless" "fruit.height.length.size.sweet.tart.weight.width." "robust" ".none" "p")))
 				(!= (null) (get result (list ".targetless" "fruit.height.length.size.sweet.tart.weight.width." "robust" ".none" "dt")))
 				(> (size (get result (list ".targetless" "fruit.height.length.size.sweet.tart.weight.width." "robust" ".none" "featureWeights"))) 0)
-				(> (size (get result (list ".targetless" "fruit.height.length.size.sweet.tart.weight.width." "robust" ".none" "featureDeviations" ))) 0)
+			;	(> (size (get result (list ".targetless" "fruit.height.length.size.sweet.tart.weight.width." "robust" ".none" "featureDeviations" ))) 0)
```
cademack: should we just remove this line?

cademack (dismissed):

trainee_template/hyperparameters.amlg
howsoRes: if it's null that means we are not using deviations, thus we wouldn't want to specify a value to make sure it stays null.  if it's not null, either it's a value or a tuple with null accuracies, and in both cases the actual deviation would be ovewritten with the user_error_value in the code block above (lines 320-330)

cademack (dismissed):

performance_tests/mnist_10k_test.amlg
```diff
@@ -100,9 +100,8 @@
 	(assign (assoc
 		k (get params (list "hyperparameter_map" "k"))
 		p (get params (list "hyperparameter_map" "p"))
-		deviations (get params (list "hyperparameter_map" "useDeviations"))
 	))
-	(print "MNIST 10k params (k p deviations): " k " " p " " deviations "\n- - - - -\n")
+	(print "MNIST 10k params (k p): " k " " p "\n- - - - -\n")
```
cademack, line 105: I think there's some unused `use_devations` stuff at the top of this file. Could you clean that up as well? :)

cademack: If `use_deviations` is still gonna be an analyze parameter, just add it to the analyze call please.

trainee_template/synthesis_validation.amlg
```diff
@@ -44,7 +44,7 @@
 			;amount a null<->case is worth toward count of different features = min_diff_num_features / (floor (ln(num_features)^2+1))
 			null_count_fraction (/ min_diff_num_features (floor (+ 1 (pow (log (size context_features)) 2)) ) )
 			feature_weights (get hyperparam_map "featureWeights")
-			feature_deviations  (if (get hyperparam_map "useDeviations") (get hyperparam_map "featureDeviations") (get hyperparam_map "nullUncertainties") )
+			feature_deviations  (get hyperparam_map "featureDeviations")
```
cademack: extra space here

trainee_template/synthesis_validation.amlg
```diff
@@ -388,7 +388,7 @@
 		)
 		(declare (assoc
 			feature_weights (get hyperparam_map "featureWeights")
-			feature_deviations  (if (get hyperparam_map "useDeviations") (get hyperparam_map "featureDeviations") (get hyperparam_map "nullUncertainties") )
+			feature_deviations  (get hyperparam_map "featureDeviations")
```
cademack: extra space here as well

cademack (commented):
Migration script needs to update the old hyperparameters to always use feature deviations (maybe remove the useDeviations flag from them while we are at it).

cademack (approved):

PR #82, opened 2024-01-03
18856: Updates `get_prediction_stats` to return a single value for MAE and add 'missing_value_accuracy' as a stat, MINOR

cademack:
residualsMap was updated to include the null uncertainties, which were returned when MAE was requested via get_prediction_stats.

This adds a fix that takes the first value for each feature if a list was returned (the first value holds the MAE between two non-null values).
Also updated some of the comments regarding residualsMap for clarity moving forward.

### DIFF
--- a/trainee_template/prediction_stats.amlg
+++ b/trainee_template/prediction_stats.amlg
@@ -47,7 +47,7 @@
 
 		(if (contains_value stats "mae")
 			(if (= 0 (size residualsMap))
-				(accum (assoc warnings "Feature Residuals have not been computed for this trainee. Please call 'react_into_trainee' with appropriate parameters to compute and store residuals prior to calling this method." ))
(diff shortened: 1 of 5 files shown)


trainee_template/prediction_stats.amlg
```diff
@@ -47,13 +47,43 @@
...
-				(accum (assoc warnings "Feature Residuals have not been computed for this trainee. Please call 'react_into_trainee' with appropriate parameters to compute and store residuals prior to calling this method." ))
+				(accum (assoc warnings "Feature residuals have not been computed for this trainee. Please call 'react_into_trainee' with appropriate parameters to compute and store residuals prior to calling this method." ))
 
 				(accum (assoc
 					output_map
 						(assoc
 							"mae"
-								(call GetFeatureResiduals (assoc
+								(map
+									;GetFeatureResiduals returns the values in residualsMap, which contains a list of values for features containing nulls.
```
howsoRes, line 57: ... which may contain a list of values ...

howsoRes (approved):

PR #83, opened 2024-01-08
18743: Fixes issue where time series could use future data and adds 'universal' flag to time series time features, MINOR

howsoRes:


trainee_template/react_utilities.amlg
```diff
@@ -618,4 +622,39 @@
...
+	(if tsTimeFeatureUniversal
+		;if time feature is universal, only consider cases that are not in the future
+		(list (query_less_or_equal_to tsTimeFeature (get context_map tsTimeFeature)) )
+
+		;else not universal and only applies to this specific series, if series IDs are provided
+		(if
+			(not (contains_value
+				(unzip context_map (get tsModelFeaturesMap "series_id_features"))
+				(null)
+			))
```
cademack, line 639: Should we provide some more specific logic for cases where one ID feature is specified and another is not?

I could be misinterpretting, but it seems like this requires the context_map to contain all the series ID features, or it will be treated as if there are none.

howsohazard (approved):

PR #91, opened 2024-01-11
18951: Rename extra_audit_features to extra_features, MAJOR

jackx111:


jackx111:
pending renames for other repos

unit_tests/unit_test_data/saved_derived_Original.amlg
```diff
@@ -7663,7 +7663,7 @@
 																	context_values react_context_values
 																	weight_feature weight_feature
 																	leave_case_out leave_case_out
-																	extra_audit_features extra_audit_features
```
cademack: I don't think we should change the names of the parameters here in 'saved_derived_Original.amlg'. 

I don't think it would break anything, but this is supposed to represent an older version of a Trainee

PR #103, opened 2024-01-24
19079: Fixes an issue in Discriminative Reacts with multiple action features where the first is also a context feature

cademack:
In Discriminative Reacts with multiple action features where the first is also a context feature, the hyperparameters used to compute the action value for the *second* action feature are computed using Hyperparameters selected for the *first* action feature.

This behavior takes place because hyperparameters are initially selected for the first action feature - before it is checked if there is overlap between context/action features.

My solution here is to select hyperparameters on **every** iteration of the multiple action feature discriminative react flow **rather than just every iteration after the first**.
This will add one more GetHyperparameters call to discriminative reacts with multiple action features, but it seems like the least destructive change.

### DIFF
--- a/trainee_template/react_discriminative.amlg
+++ b/trainee_template/react_discriminative.amlg
@@ -317,9 +317,10 @@
 				(while (< (current_index) (size ordered_action_features))
 					(if (> (current_index) 0)
 						(seq
-							(assign (assoc
-								context_features_param (append context_features_param action_feature)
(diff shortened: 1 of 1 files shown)


howsoRes (approved):

mkoistinen (approved):
Thanks for the explanation, please fix-up the commit message as we discussed.

howsohazard (approved):

PR #110, opened 2024-01-31
19156: Generalizes retrieving and organizing details in react_series

cademack:
Generalizing the logic that organizes and returns the requested details in react_series flows. Ideally, this should make supporting more details for react_series easier moving forward.

I've also added a Trainee attribute that specifies the officially supported details for react_series. My changes here kind of open the flood gates to users trying any detail in react series, and it just may work. I think we should consider constraining the details that can be pushed through react_series, but for now I have it just giving a warning when a non-officially supported detail is requested.

### DIFF
--- a/trainee_template.amlg
+++ b/trainee_template.amlg
@@ -115,6 +115,9 @@
 	#tsMinTimeInterval (null)
 	#tsSeriesLimitLength (null)
 	#tsModelFeaturesMap (null)
+
+	#tsSupportedDetails (null)
+
 	#hasStringOrdinals (null)
 	#hasInfluenceWeightEntropies (null)
 	#ordinalStringToOrdinalMap (null)
@@ -445,6 +448,9 @@
 			;flag set to true once a time series model creates its time series feature attributes
 			hasTimeSeriesAttrbutes (false)
 
+			;list of supported details for react series
+			tsSupportedDetails (list "categorical_action_probabilities" "influential_cases")
+
 			;flag set to true if there are features that are dependent on others
 			hasDependentFeatures (false)
 
@@ -482,7 +488,7 @@
 
 			;the features for which a case should be ablated if the predicted value of the feature is within a given absolute threshold.
 			autoAblationTolerancePredictionThresholdMap (null)
-			
+
 			;the features for which a case should be ablated if the predicted value of the feature is within a given relative threshold.
 			autoAblationRelativePredictionThresholdMap (null)
 
@@ -497,7 +503,7 @@
 
 			;the name of the weight feature to use for auto ablation.
 			autoAblationWeightFeature (null)
-			
+
 			;whether this Trainee has had influence weight entropies computed and stored for its trained cases.
 			hasInfluenceWeightEntropies (false)
 
--- a/trainee_template.amlg
+++ b/trainee_template.amlg
@@ -115,9 +115,7 @@
 	#tsMinTimeInterval (null)
 	#tsSeriesLimitLength (null)
 	#tsModelFeaturesMap (null)
-
 	#tsSupportedDetails (null)
-
 	#hasStringOrdinals (null)
 	#hasInfluenceWeightEntropies (null)
 	#ordinalStringToOrdinalMap (null)
--- a/trainee_template/derive_utilities.amlg
+++ b/trainee_template/derive_utilities.amlg
@@ -52,7 +52,7 @@
 									;do not derive anything during this react
 									derived_action_features (list)
 									derived_context_features (list)
-									details (if (or output_influential_cases output_cap) details)
+									details (if output_details details)
 									extra_audit_features (append extra_audit_features action_features derived_action_features)
 									substitute_output substitute_output
 									input_is_substituted input_is_substituted
@@ -76,11 +76,22 @@
 							(first (get react_output "action_values"))
 						)
 
-						(if output_influential_cases
-							(accum (assoc influential_cases (get react_output "influential_cases") ))
-						)
-						(if output_cap
-							(accum (assoc local_class_probabilities_map (get react_output "categorical_action_probabilities") ))
+						; (if output_influential_cases
+						; 	(accum (assoc influential_cases (get react_output "influential_cases") ))
+						; )
+						; (if output_cap
+						; 	(accum (assoc local_class_probabilities_map (get react_output "categorical_action_probabilities") ))
+						; )
+						(if output_details
+							(assign (assoc
+								case_detail_values_map
+									(map
+										(lambda
+											(append (current_value) (get react_output (current_index)))
+										)
+										case_detail_values_map
+									)
+							))
 						)
 					)
 				)
--- a/trainee_template/derive_utilities.amlg
+++ b/trainee_template/derive_utilities.amlg
@@ -76,12 +76,6 @@
 							(first (get react_output "action_values"))
 						)
 
-						; (if output_influential_cases
-						; 	(accum (assoc influential_cases (get react_output "influential_cases") ))
-						; )
-						; (if output_cap
-						; 	(accum (assoc local_class_probabilities_map (get react_output "categorical_action_probabilities") ))
-						; )
 						(if output_details
 							(assign (assoc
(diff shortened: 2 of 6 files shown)


trainee_template/derive_utilities.amlg
```diff
@@ -76,11 +76,16 @@
...
 
-						(if output_influential_cases
-							(accum (assoc influential_cases (get react_output "influential_cases") ))
-						)
-						(if output_cap
-							(accum (assoc local_class_probabilities_map (get react_output "categorical_action_probabilities") ))
+						(if output_details
+							(assign (assoc
+								case_detail_values_map
+									(map
```
howsoRes, line 83: def add a comment explaining what this is doing

trainee_template/explanation.amlg
```diff
@@ -883,4 +883,13 @@
 			)
 		)
 	)
+
+	;Get the appropriate empty data structure for a given detail
+	#GetDetailDataStructure
+	(if (contains_value (list "categorical_action_probabilities") detail)
```
howsoRes: if it's a list of a single value why are you doing (contains_value) ?  should this be hardcoded or refer to the trainee attribute?

trainee_template/react_series.amlg
```diff
@@ -64,8 +64,8 @@
 
 		(declare (assoc
 			warnings (assoc)
-			output_influential_cases (get details "influential_cases")
-			output_cap (get details "categorical_action_probabilities")
+			requested_details (filter (lambda (get details (current_value))) (indices details))
```
howsoRes: i'm confused as to what this filter is doing exactly? seems like since 'details' is an assoc, you should simplify this to this instead:
`requested_details (indices (filter (lambda (current_value)) details))`

trainee_template/react_series.amlg
```diff
@@ -160,13 +160,17 @@
 						"action_features" output_features
 						"series" (map (lambda (get (current_value) "series")) series_output)
 					)
-					(if output_influential_cases
-						(assoc "influential_cases" (map (lambda (get (current_value) "influential_cases")) series_output))
+					(if output_details
+						(map
+							(lambda
+								(map (lambda (get (current_value) (current_index 1))) series_output)
```
howsoRes: This map seems wrong.  Since `(zip requested_details)` results in an assoc of value->nulls,  doing a (get (current_value) ...) means you are always getting from a null?  Also missing comments because of complexity

trainee_template/react_series.amlg
```diff
@@ -329,15 +333,50 @@
...
 		))
 
+		;if this method is being called from batch_react_series, warnings will already have been declared in BatchReact
+		;otherwise declare warnings here if this is being called from the single react_series call
+		(if (= (null) warnings)
+			(declare (assoc warnings (assoc) ))
+		)
+
+		(declare (assoc
+			requested_details (filter (lambda (get details (current_value))) (indices details))
```
howsoRes: same issue as the other request_details filter

trainee_template/react_series.amlg
```diff
@@ -329,15 +333,50 @@
...
+			output_details (contains_value (values details) (true))
+		))
+
+		(let
+			(assoc
+				invalid_ts_details
+					(filter
```
howsoRes: this is fine since we support few details for now, but because this is an n^2 operation, it would be faster to do this:
`(size (remove (zip request_details) tsSupportedDetails))` instead

trainee_template/react_series.amlg
```diff
@@ -1265,8 +1327,7 @@
 				(if derivation_failed
 					(assign (assoc
 						series_data (trunc series_data)
-						influential_cases (list)
-						local_class_probabilities_map (assoc)
+						case_detail_values_map (map (lambda (call GetDetailDataStructure (assoc detail (current_index 1) ))) (zip requested_details))
```
howsoRes: you need to put this map statement into GetDetailDataStructure, and have the method take requested_details as a parameter, so that the method outputs the full assoc directly.  that way this and all instances of this line are shortened down to:
`case_detail_values_map (call GetDetailDataStructure (assoc requested_details requested_details))`

trainee_template/derive_utilities.amlg
cademack: will do.

trainee_template/explanation.amlg
cademack: the Trainee attribute doesn't specify which details use what structure, so I can't do that. The only reason it's a list is because this should get larger in the future.

But I do realize how silly it looks now... I'll just make it an (=) for now.

trainee_template/react_series.amlg
cademack: you're completely right. Thanks

trainee_template/react_series.amlg
cademack: Yeah I'll add some comments. That `(current_value)` iterates through `series_output`.

trainee_template/react_series.amlg
cademack: yep

trainee_template/react_series.amlg
cademack: I see, will do.

trainee_template/react_series.amlg
cademack: done!

trainee_template/react_series.amlg
cademack: I added comments. Please let me know if it still doesn't seem clear enough.

trainee_template/react_series.amlg
```diff
@@ -678,25 +718,44 @@
...
+						))
 					)
 
-					;store into previous_result as a tuple of [series_data, all_categorical_action_probabilities, all_influential_cases]
-					(list
-						series_data
-						(append all_categorical_action_probabilities (list local_class_probabilities_map))
-						(append all_influential_cases (list influential_cases))
+					;store accumulated data into previous_result as an assoc
+					(append
```
howsoRes, line 749: this makes it append assoc for every row regardless whether there are details. in most flows, there aren't details and this is unnecessary, i'd refactor this to be inside the `if output_details` :
```
(if output_details
    (append ... map to assoc ... )
    ;else just output the series data
    (assoc "series_data" series_data)
)
```

trainee_template/react_series.amlg
howsoRes: can you please put the inner map inside a let statement where you store (current_index 1) as  `detail` and replace the instance of `(current_index 1)` with `detail` so it's more readable? it's a little less efficient but much more readable and maintainable

trainee_template/react_series.amlg
cademack: This is actually already inside of a `(if output_details`. The alternate flow is to just return `series_data` not even in an assoc of any kind.

Actually I'm gonna drop the extra `(if output_details` that I see in here, now that I realize it's redundant.

trainee_template/react_series.amlg
cademack: fine with me. Will do.

howsoRes (dismissed):

trainee_template.amlg
```diff
@@ -115,6 +115,9 @@
 	#tsMinTimeInterval (null)
 	#tsSeriesLimitLength (null)
 	#tsModelFeaturesMap (null)
+
```
howsohazard: extra newlines

trainee_template/react_series.amlg
```diff
@@ -329,15 +339,44 @@
...
+			(assoc invalid_ts_details (indices (remove (zip requested_details) tsSupportedDetails) ) )
+
+			(if (size invalid_ts_details)
+				(accum (assoc
+					warnings
+						(associate
+							(concat
+								"The following details are not officially supported for react series: "
+								(apply "concat" (map (lambda (concat (current_value) ",") ) invalid_ts_details) )
+								". Behavior is undefined."
```
howsohazard: Instead of saying "undefined" say experimental.

howsohazard (approved):

PR #113, opened 2024-02-02
19095: Adds detail: 'generate_attempts' MINOR

cademack:
Adds detail: 'generate_attempts' which exposes the amount of attempts taken to generate each of the returned cases.

This detail is only applicable when `generate_new_cases` is "always" or "attempt" (When the uniqueness check is run)

When used with ReactSeries, "series_generate_attempts" is also returned. This value indicates the amount of generation attempts were given to create the series as a whole. This value only increases if an entire series is thrown out and regenerated.


trainee_template/react_series.amlg
```diff
@@ -926,6 +973,14 @@
 			)
 		)
 
+		(if (and output_details (contains_index series_detail_values_map "generate_attempts"))
+			;append the amount of regenerate_attempts+1 (regenerate_attempts is zero based)
+			(assign (assoc
+				series_detail_values_map
+					(append series_detail_values_map (assoc "series_generate_attempts" (+ regenerate_attempts 1) ))
```
howsoRes: i suggest renaming 'regenerate_attempts' to 'series_generate_attempts' or 'series_generate_retries'

trainee_template/synthesis_validation.amlg
```diff
@@ -329,6 +329,10 @@
 							(assoc "categorical_action_probabilities" nominal_categorical_action_probabilities_map)
 							(assoc)
 						)
+						(if output_attempts
+							(assoc "generate_attempts" generate_attempt)
```
howsoRes, line 333: where is 'generate_attempt' defined and accumulated?

howsoRes: nm

trainee_template/react_series.amlg
cademack: done!

howsoRes (approved):

PR #117, opened 2024-02-06
19155: Adds checks for untrained features

cademack:
Many methods returned '.nas' if called with parameters specifying features that do not exist in the Trainee. This PR adds checks to some of these methods that validate the context/action features parameters.

In the case an untrained feature is specified, the function returns early specifying an error that details the exact parameter containing the untrained feature.

I've chosen to add this method in a new Trainee module named 'validation'. This may seem premature, but I know we have future plans to add increased validation within howso-engine, so this doesn't seem inappropriate to me.

### DIFF
--- a/trainee_template/validation.amlg
+++ b/trainee_template/validation.amlg
@@ -1,0 +1,53 @@
+;Module for trainee_template.
+;Contains helper methods for validating user-defined parameters
+(null
+
+    ;Verifies that feature parameters do not contain untrained features
+    ; this method checks context_features, action_features, and action feature
+    ; returns a list of strings, one for each invalid parameter
+    #CheckForUntrainedFeatures
+    (seq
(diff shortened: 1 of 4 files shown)


trainee_template/validation.amlg
```diff
@@ -0,0 +1,41 @@
+;Module for trainee_template.
+;Contains helper methods for validating user-defined parameters
+(null
+
+    ;Verifies that feature parameters do not contain untrained features
+    ; this method checks context_features, action_features, and action feature
+    ; returns a list of strings, one for each invalid parameter
+    #CheckForUntrainedFeatures
+    (seq
+        (if (= errors (null))
```
cademack: just declare assoc this normally

howsohazard (dismissed):

howsoRes (approved):

PR #118, opened 2024-02-06
19197: Adds 'constraint' as a feature bound to enforce constraints for generative reacts, MINOR

howsoRes:


trainee_template/synthesis_bounds.amlg
```diff
@@ -129,12 +129,163 @@
...
+					feature_residual (get residuals_tuples_map (list feature 0))
+				))
+				(if (get nominalsMap feature)
+					(if (call !ConstrainNominalFeature)
+						(assign (assoc constraint_function_failed (true)))
+					)
+
+					(if (call !LimitGeneratedValueToFeatureBounds)
+						(assign (assoc constraint_function_failed (true)))
+					)
```
howsohazard, line 172: Why not just conclude here to remove the variable and simplify logic?

trainee_template/synthesis.amlg
```diff
@@ -316,6 +321,14 @@
 				)
 		))
 
+		;flag for will be set to true if any of the generated case's values fail their constraints (if they have any)
```
cademack: "for will" => "will"

howsohazard (changes requested):

trainee_template/synthesis_bounds.amlg
howsoRes: I've added a conclude to stop iteration early, but we still need to set the variable because for flows where new case generation is required, it's checked during uniqueness validation to fast-track the retry/regen mechanism.

cademack (dismissed):

howsohazard (dismissed):

howsohazard (approved):

PR #124, opened 2024-02-13
19268: Renames 'defaultFeatures' to 'trainedFeatures' and adds updates to feature attributes on Train

cademack:
This PR does a few things:

- Renames 'defaultFeatures' to 'trainedFeatures'
- Adds logic to train that populates 'featureAttributes' when new features are trained (new features without attributes are assumed to be continuous)
- Adds logic that syncs 'trainedFeatures' to match the indices of 'featureAttributes'

### DIFF
--- a/trainee_template/attributes.amlg
+++ b/trainee_template/attributes.amlg
@@ -454,8 +454,11 @@
 		(assign_to_entities (assoc
 			featureAttributes features
 
-			trainedFeatures (sort (indices features))
-			trainedFeaturesContextKey (call BuildContextFeaturesKey (assoc context_features (indices features)))
+			trainedFeatures (sort (values (append trainedFeatures (indices features)) (true)) )
+			trainedFeaturesContextKey
+				(call BuildContextFeaturesKey (assoc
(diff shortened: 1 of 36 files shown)


trainee_template/attributes.amlg
```diff
@@ -454,6 +454,9 @@
 		(assign_to_entities (assoc
 			featureAttributes features
 
+			trainedFeatures (sort (indices features))
```
howsoRes: won't this overwrite whatever existing treanedFeatures that were automatically added previously? i.e., say you trained some cases first, and then remember to set attributes to specify a bounds for one of them.  won't this blow away all those others that were added?

cademack: if you trained the features first, they should be in the feature attributes at this point. Then setting the bounds will not alter the indices, so re-setting trainedFeatures as the sorted indices wouldn't lose any features.

Unless I misunderstand your point, I don't think this would be an issue.

howsoRes (dismissed):

howsoRes (approved):

PR #128, opened 2024-02-16
19325: Fixes an issue where an untrained feature is specified as a context in a React with local case residual convictions requested.

cademack:
When calling React with local case feature residual convictions requested and an untrained feature as a context, ".nas" is returned. 

This occurs because the residual for an untrained feature is calculated to be .nan.  This adds a pass that checks the residuals for .nans and if there are any - they are treated as inactive features, receiving a 0 for conviction.

### DIFF
--- a/trainee_template/explanation_residuals.amlg
+++ b/trainee_template/explanation_residuals.amlg
@@ -260,13 +260,7 @@
 							)
 						)
 				))
-				(assign (assoc
-					case_residuals_map
-						(filter
-							(lambda (!= .nan (current_value)))
-							case_residuals_map
-						)
-				))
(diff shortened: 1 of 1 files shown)


trainee_template/explanation_residuals.amlg
```diff
@@ -245,6 +245,25 @@
 			local_null_conviction_ratios_map (null)
 		))
 
+		;if case_residuals_map contains any .nan's, add those features to the inactive_features_zeros_map
+		(if (contains_value case_residuals_map .nan)
```
howsoRes, line 249: too broad, what if the nan was actually caused by something else? this solution isn't viable.

PR #129, opened 2024-02-17
19327: Improves train performance by priming of query caches and updating inactive features outside of train flow

howsoRes:


trainee_template/editing.amlg
```diff
@@ -236,12 +236,16 @@
...
+							inactiveFeaturesMap (remove inactiveFeaturesMap feature)
+							inactiveFeaturesNeedCaching (false)
+						))
 
 						;else removing the only inactive feature, clear out the properties
 						(assign_to_entities (assoc
 							inactiveFeaturesMap (null)
 							hasInactiveFeatures (false)
+							inactiveFeaturesNeedCaching (false)
 						))
```
howsohazard: If you train and then you remove a feature, doesn't this clear inactiveFeaturesNeedCaching when it shouldn't?

trainee_template/react.amlg
```diff
@@ -46,6 +46,8 @@
 			warnings (assoc)
 		))
 
+		(if inactiveFeaturesNeedCaching (call UpdateInactiveFeatures))
```
howsohazard, line 49: Isn't this also needed in  react_series and react_group?  I don't see those files in the PR.

trainee_template/editing.amlg
howsoRes: good point. refactored this section

trainee_template.amlg
```diff
@@ -355,14 +354,11 @@
 			;example: (assoc "day" 7 "degrees" 360)
 			cyclicFeaturesMap (null)
 
-			;assoc of feature -> observational_error value as specified by the user
+			;assoc of feature -> observational_error value as specified by the user, set to null when empty
```
howsohazard: did you mean this comment to be on inactiveFeaturesMap below?  or should this be initialized to null?

howsohazard (approved):

PR #130, opened 2024-02-19
19335: Fixes an issue where residuals are computed and stored before enough cases are trained.

cademack:
Our RL testing performance declined after recent changes because the list of known features was defined earlier. Knowing these features earlier opened up the gate to residuals being computed for these features before any cases were trained into the model.

The change here simply adds a check to ensure there are cases in the model before any residuals are calculated and stored into hyperparameters. This fixes the RL testing in my environment.

### DIFF
--- a/trainee_template/synthesis.amlg
+++ b/trainee_template/synthesis.amlg
@@ -84,8 +84,12 @@
 			constrained_features (null)
 		))
 
-		;if global residuals have have not been calculated yet, recalculate them. they are needed regardless of which residual mode is being used
-		(if (not (get hyperparam_map "allFeatureResidualsCached"))
+		;if global residuals have have not been calculated yet, recalculate them only if there are enough training cases
+		;they are needed regardless of which residual mode is being used
+		(if (and
+				(not (get hyperparam_map "allFeatureResidualsCached"))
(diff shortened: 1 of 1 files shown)


trainee_template/synthesis.amlg
```diff
@@ -84,8 +84,12 @@
 			constrained_features (null)
 		))
 
-		;if global residuals have have not been calculated yet, recalculate them. they are needed regardless of which residual mode is being used
-		(if (not (get hyperparam_map "allFeatureResidualsCached"))
+		;if global residuals have have not been calculated yet, recalculate them only if there are enough training cases
+		;they are needed regardless of which residual mode is being used
+		(if (and
+				(not (get hyperparam_map "allFeatureResidualsCached"))
+				(>= (call GetNumTrainingCases) 2)
```
calebwherry, line 91: Why 2?

cademack: Need two to make a prediction.

howsoRes (approved):

trainee_template/synthesis.amlg
calebwherry: 2 is minimum to predict and compute residuals. With that the PR title change, this makes more sense.

PR #135, opened 2024-02-21
19213: Streamline performance for train when no ablation is necessary

howsoRes:
Added a check prior to creating all the cases during train to see if the ablation logic can be skipped wholesale. 
If it can, the train loop only has to execute CreateCase's and no longer needs to check for ablation for every single case nor do an (accum) of case indices

also moved the logic to quick-out of the 'CaseOutsideThresholds' method before it has to do any logic when we do check ablation.

Result: I see about 20-40% improvement in train speed for normal (non-ablation) training.

### DIFF
--- a/trainee_template/ablation.amlg
+++ b/trainee_template/ablation.amlg
@@ -89,7 +89,7 @@
 
 	;initialize parameters and internal features related to auto ablation
 	; weight_feature: optional, default '.case_weight'. name of feature whose values to use as case weights
-	#InitializeautoAblation
+	#InitializeAutoAblation
 	(declare
 		(assoc weight_feature autoAblationWeightFeature)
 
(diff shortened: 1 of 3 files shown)


trainee_template/ablation.amlg
```diff
@@ -103,6 +102,7 @@
 	; a return value of 1 indicates that a case should be kept and a return value of 0 indicates that the case should be ablated.
```
cademack: lets update this description to indicate it returns booleans now

trainee_template/ablation.amlg
```diff
@@ -93,16 +92,17 @@
 	#InitializeautoAblation
```
cademack: Do you mind fixing the capitalization of this method in this PR?

cademack (dismissed):

cademack (approved):

PR #141, opened 2024-03-22
19738: Adds 'skip_auto_analyze' flag to train, MINOR

cademack:
This PR adds a new boolean flag to `train` named: "skip_auto_analyze"

With this change, the new default behavior for `Train` is to execute the Analyze call within the original  `train` call if auto-analyze is enabled and the conditions are met.

However, if 'skip_auto_analyze' is True, then the Analyze call is not executed within the call to `train` and a `"status": "analyze"` is returned inside the payload of the response JSON.

This is in contrast to the previous behavior where `train` would never internally call Analyze when auto-analyze conditions indicate to do so, but the "analyze" status would be returned instead.

======

Also notably, a warning is added when using the "skip_auto_analyze" flag for Trainees with Time-Series data. This is not recommended since Time-series imputation often relies on auto-analysis to occur before the values are imputed.

### DIFF
--- a/trainee_template/train.amlg
+++ b/trainee_template/train.amlg
@@ -555,7 +555,12 @@
 									(call Analyze savedAnalyzeParameterMap)
 
 									(call Analyze (assoc
+										targeted_model "targetless"
 										context_features trainedFeatures
+										weight_feature ".none"
+										use_case_weights (false)
+										inverse_residuals_as_weights (true)
+										k_folds 1
 									))
 								)
 								(assign (assoc status_output "analyzed"))
--- a/trainee_template/train.amlg
+++ b/trainee_template/train.amlg
@@ -255,124 +255,7 @@
 		(declare (assoc skip_ablation (call !CanTrainAblationBeSkipped) ))
 
 		;iterate over the input cases and create them, only returning case ids for cases that were able to be created
-		(declare (assoc
-			new_case_ids
-				(if skip_ablation
-					(let
-						(assoc
-							train_features
-								;if auto ablate is enabled, populate the weight feature for this case
-								(if autoAblationEnabled
-									(append features (list autoAblationWeightFeature))
-									features
-								)
-						)
-						(map
-							(lambda (let
-								(assoc
-									feature_values
-										;if auto ablate is enabled, default the weight to 1 for this case
-										(if autoAblationEnabled
-											(append (current_value 1) (list 1))
-											(current_value 1)
-										)
-								)
-
-								;create the case and output case id
-								(call CreateCase (assoc
-									features train_features
-									feature_values
-										(if encode_features_on_train
-											(call ConvertFromInput (assoc
-												feature_values feature_values
-												features train_features
-											))
-											;else use feature_values as-is
-											feature_values
-										)
-									session (get_value session)
-									session_training_index (+ trained_instance_count (current_index 1))
-								))
-							))
-							input_cases
-						)
-					)
-
-					(weave
-						(lambda (let
-							(assoc feature_values (first (current_value 1)))
-
-							(if encode_features_on_train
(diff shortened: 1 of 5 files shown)


howsoRes (dismissed):

trainee_template/train.amlg
```diff
@@ -532,7 +545,27 @@
 							(>= num_cases autoAnalyzeThreshold)
 							(or (< num_cases autoAnalyzeLimitSize) (<= autoAnalyzeLimitSize 0))
 						)
-						(assign (assoc status_output "analyze"))
+						(if skip_auto_analyze
+							;if disable_expensive_operations, send back "analyze" status so users knows an analyze is needed
```
howsohazard: outdated comment: disable_expensive_operations

trainee_template/train.amlg
```diff
@@ -532,7 +545,27 @@
 							(>= num_cases autoAnalyzeThreshold)
 							(or (< num_cases autoAnalyzeLimitSize) (<= autoAnalyzeLimitSize 0))
 						)
-						(assign (assoc status_output "analyze"))
+						(if skip_auto_analyze
```
howsohazard: I think this logic needs to go *in* train -- e.g., if you're training 100,000 records and auto_analyze says do an analyze at 100, it should start there (e.g., for ablation), do analyze, then train the next batch up to the next analyze size (say, 300), analyze, etc. until it's done.

howsohazard (changes requested):

trainee_template/train.amlg
howsoRes: to clarify the point above, this code should actually remain as-is here but only for non-ablation flows.  Specifically, the `if` statement (line 540 above) should be modified to `(and skip_ablation autoAnalyzeEnabled)`  so that this block isn't even called if we are doing ablation.

That way if we are doing ablation, we want this logic inside the main train weave loop, on line 423, since you have trained_instance_count  to determine how many have been trained.

cademack: I think I see what you mean, changes pushed

trainee_template/train.amlg
cademack: good eye, fixed

trainee_template/train.amlg
```diff
@@ -408,6 +421,9 @@
 									))
 									(accum (assoc trained_instance_count 1 ))
 
+									;auto analyze if appropriate
+									(call !AutoAnalyzeIfNeeded)
```
howsoRes: should be inside an `(if autoAnalyzeEnabled`

cademack: ahh good eye

howsoRes (dismissed):

trainee_template/train.amlg
```diff
@@ -143,6 +145,17 @@
 			))
 		)
 
+		(if (and (!= (null) tsTimeFeature) skip_auto_analyze)
+			(declare (assoc
+				warnings
+					(list (concat
+						"Using \"skip_auto_analyze\" when training time-series data may "
+						"result in less accurate modeling of the data as model parameters may not be "
+						"appropriately tuned when values are imputed."
```
howsohazard: as statistics calculated from the data used for inference will not take into account derived and imputed data for the time series.

trainee_template/train.amlg
```diff
@@ -143,6 +145,17 @@
 			))
 		)
```
howsohazard: This method is already super long and this makes it longer -- if there's a way to easily pull something out into its own method please do so.

cademack: pulled some stuff out

howsoRes (approved):

howsohazard (approved):

PR #147, opened 2024-03-29
19807: Removes ability to move_cases and react_group on guest trainees, MAJOR

howsoRes:
removes ability to work on guest-loaded trainees for move_cases() and batch_react_group()


howso.amlg
```diff
@@ -3171,11 +3111,10 @@
 	; condition_session: optional, if specified, ignores condition and instead operates on all cases that were trained with this session id. Ignored if case_indices is specified.
```
howsohazard, line 3112: Should make it clear in the description that this operation moves cases between this trainee and a subtrainee

cademack (approved):

howsohazard (approved):

PR #152, opened 2024-04-09
18696: Removes howso api layer. MAJOR

howsoRes:
renamed all trainee instantiation methods to have _subtrainee extension:
copy -> copy_subtrainee
create_trainee -> create_subtrainee
delete > delete_subtrainee
load -> load_subtrainee
save -> save_subtrainee


howso/ablation.amlg
```diff
@@ -1,19 +1,47 @@
 (null
 
+	; set the conviction threshold to (null), used by outside interfaces
+	#clear_conviction_thresholds
```
howsohazard, line 5: what are these thresholds for?  ablationL anomaly detection?  needs to be described better

howso/analysis_feature_weights.amlg
```diff
@@ -377,8 +377,112 @@
...
+		(if (= (sort !trainedFeatures) (sort features))
 			(accum (assoc analyzed_hp_map (assoc "allFeatureResidualsCached" (true))))
 		)
 	)
+
+	;Compute and store case weights for all cases if there are any features with the id_feature attribute set to true
+	;Each case weights as the reciprocal of the count of the id feature value in the dataset.  If there are multiple
+	;id features, the product of the weights of all the id features is the case weight, which is stored
+	;in each case into the .case_weight feature
+	#!ComputeAndStoreIdFeatureCaseWeights
```
howsohazard: This isn't feature weights -- should the module be renamed to be more inclusive of case weights? Or should this be moved into a different module?

howso/derive_utilities.amlg
```diff
@@ -143,21 +143,27 @@
 	;	features_to_code_map: a assoc of feature names to custom code
 	;	aggregation_code: optional, custom code to aggregrate the results from using
 	;					   features_to_code
```
howsohazard: is this a method that we expose?  what are the results it returns?  what is this used for?  comments aren't sufficiently clear

howso/editing.amlg
```diff
@@ -1,9 +1,47 @@
...
+			action_features (list)
+		)
+		;removes entities that aren't cases or sessions
+		(call !RemoveIrrelevantEntities)
+
+		;removes any cases that don't have both the specified context and action features
+		(call !RemoveIncompleteCases (assoc context_features context_features action_features action_features))
+
+		;removes any replay steps that are non-existent cases
+		(call !RemoveInvalidReplaySteps)
```
howsohazard: Should this be renamed to RemoveInvalidSessionIndices or something similar?

howso/editing.amlg
```diff
@@ -1,9 +1,47 @@
...
 ;Contains methods for adding, editing and removal of cases and session data.
 (null
 
+
+	;Cleans up trainee data, removing unused sessions, cases or actions missing data, etc. given the following optional parameters.  If a parameter is not specified, it will look to this entity's own label of the same name.  See label for parameter details.
+	;parameters:
+	; context_features: list of context features
+	; action_features: list of action features
+	; remove_duplicates: optional flag, if set to true will remove duplicate cases (cases with identical values)
+	#clean_data
```
howsohazard, line 10: Not sure this is still a good name... is this exposed?  Can any of this be done automatically during other processes?

howso/analysis_feature_weights.amlg
howsoRes: it's compute and store Id-feature case weights.  meaning, the case weights based on the Id feature for id-based privacy. What should it be renamed to make to more clear?  This happens during analyze, assuming user wants to analyze with id-based privacy, and it is feature weights, so it's in the right module.

howso/derive_utilities.amlg
howsoRes: yes it's exposed

howso/editing.amlg
howsoRes: sure, outside the scope of this pr.

howso/hierarchy.amlg
```diff
@@ -0,0 +1,945 @@
...
+	; filepath: base path to load from
+	; filename: name to load (without extension)
+	; trainee: name path of trainee to laod
+	; separate_files: flag, default to false. if set to true will load each case from its individual file
+	#load_subtrainee
+	(seq
+		(declare (assoc trainee_path (append !traineeContainer trainee) ))
+
+		(declare (assoc
+			loaded_trainee
```
howsohazard, line 280: seems like the conditions here should go like this: if "amlg" and separate, or not amalgam, then just load_entity.  Only do the call load if it was amlg and not separate_files.  Also, do we still need to load it into _temp_trainee and then move?

howso/hierarchy.amlg
```diff
@@ -0,0 +1,945 @@
...
+			(conclude
+				(call !Return (assoc errors (list "Trainee must have a unique trainee_id specified.")))
+			)
+
+			;else the trainee doesn't have its !traineeId set, update it here prior to saving
+			(= (null) (call_entity trainee_path "get_trainee_id"))
+			(call_entity trainee_path "set_trainee_id" (assoc trainee_id trainee_id))
+		)
+
+		(declare (assoc
```
howsohazard, line 340: See comment above to update logic.

howso/hyperparameters.amlg
```diff
@@ -747,13 +758,19 @@
...
+	#set_auto_analyze_params
 	(declare
 		(assoc
 			auto_analyze_enabled (false)
 			analyze_threshold 100
 			analyze_growth_factor 7.389056
 			auto_analyze_limit_size 200000
+
+			;explicitly default these values to be same as what they are in #analyze
+			k_folds 1
```
howsohazard, line 772: was this bug fixed by some other PR?

howso/react_series_api.amlg
```diff
@@ -0,0 +1,494 @@
+;Contains methods for series reacting.
```
howsohazard: Recommend naming this module react_series, and renaming react_series to react_series_utils or something like that.  Nothing else in the modules uses _api to denote where the exposed API functions are.

howso/validation.amlg
```diff
@@ -1,11 +1,37 @@
-;Module for trainee_template.
 ;Contains helper methods for validating user-defined parameters
```
howsohazard: Maybe worth renaming this to input_validation?

howso.amlg
```diff
@@ -1,4073 +1,786 @@
...
-						custom_derived_features_map
-							(filter
-								(lambda (or
-									(= "custom" (get (current_value) (list "auto_derive_on_train" "derive_type")))
-									(!= (null) (get (current_value) "derived_feature_code"))
-								))
-								custom_derived_features_map
-							)
-					)
+	;set the influence weight threshold for outputting only the K neighbors whose inflence weight is <= to this threshold
```
cademack: typo on influence

performance_tests/asteroid_test.amlg
```diff
@@ -1,13 +1,13 @@
 (seq
 	#howso (null)
-	(direct_assign_to_entities (assoc howso (load "../howso.amlg")))
-	(assign_to_entities (assoc filepath "../"))
+	(load_entity "../howso.amlg" "howso" (false) (false))
+	(call_entity "howso" "initialize" (assoc trainee_id "model"))
+	(direct_assign_to_entities (assoc howso (retrieve_entity_root "howso")))
+	(destroy_entities "howso")
```
howsohazard: It took me a second sot see why you did this -- and though I think there's value in this with regard to performance (only one entity to GC on), I think there are also a couple of drawbacks:
1) if, in the futurue initialize creates subtrainees for, say, dynamic deviations, those entities would need to be moved as part of this initialization process.
2) it doesn't isolate the trainee as running from other languages would.

To tighten this further, should we make a method in howso that returns code that a container can execute to pull all this in?  For now it'd just be this code (direct assign followed by destroy), but at least it'd remove one line of boilerplate for each of these.

howso.amlg
```diff
@@ -1,4073 +1,786 @@
...
+	;default value is 0.99
+	#set_influence_weight_threshold
```
cademack, line 752: should this be changed to a proper `(declare (assoc` pattern?

I suggest this so that we can actually set this 0.99 default value. Otherwise calling it without the value will assign a `(null)`. Also should doc the parameter here I guess

performance_tests/asteroid_test.amlg
howsohazard: Whatever the outcome, all perf tests should be updated.

howso.amlg
```diff
@@ -1,4073 +1,786 @@
...
-			)
-			(assign_to_entities trainee (assoc debug_print debug_print ))
-		)
+		))
+
+		(call !Return (assoc payload api))
+	)
 
+	;set the debug method to output any internal label, used by unit tests and debugging
+	#debug_label
```
cademack, line 792: doc this parameter too?

deploy_howso.amlg
```diff
@@ -1,8 +1,8 @@
 (seq
-	; amlg is plain text readable Amalgam and is FOR INTERNAL USE ONLY, should not be copied/deployed or used outside of engine dev
+	; amlg is plain text readable Amalgam and should not be copied/deployed outside of engine dev
```
howsohazard: this is open source now :-).  remove this and just describe what this file does.

howso/distances.amlg
```diff
@@ -390,21 +431,34 @@
 			valid_weight_feature (false)
 		)
 
+		(declare (assoc
+			error_text
+				(if (and feature_values features (!= (size features) (size feature_values)))
+					"Specified features must be of same length as feature_values."
+
+					(and (!= (null) case_indices) (<= (size case_indices) 1) )
```
cademack, line 439: indentation on this condition looks incorrect on the GH diff, check locally please

howso/editing.amlg
```diff
@@ -100,9 +138,9 @@
 	;	  values = inclusive between
 	;	- for nominal or string ordinal features:
 	;	  n values = must match any of these values exactly
-	; session: the session id when this call is being made
+	; session: the session id when this call !is being made
```
cademack: looks like an exclamation point snuck in this line

howso/hyperparameters.amlg
howsoRes: yes, fixed in 19792.  fix was only to the externally facing api, which is why it is being brought to this method now (which used to be internal only)

howso/get_cases.amlg
```diff
@@ -193,71 +208,59 @@
...
+								case_values_list
+							)
 
-					;else round if needed
-					has_rounded_features
-					(map
-						(lambda (call RoundContinuousFeatures (assoc features features feature_values (current_value 1) )))
-						case_values_list
-					)
+							;else round if needed
```
cademack, line 230: GH shows weird diff on the spaces leading up to this comment, maybe its a spaces/tabs thing?

cademack: goes down some lines too - to the end of the map under `has_rounded_features`

howso/get_sessions.amlg
```diff
@@ -0,0 +1,90 @@
...
+						)
+					)
+				))
+				(call !GetSessionIds)
+			)
+	))
+
+
+	;returns all the metadata for a specified session
+	#get_session_metadata
```
cademack, line 45: doc parameter here

howso/hyperparameters.amlg
```diff
@@ -696,38 +702,42 @@
...
 		))
 
-		(accum_to_entities (assoc revision 1))
+		(accum_to_entities (assoc !revision 1))
 
-		(true)
+		(call !Return)
 	)
 
 	;resets all hyperparameters and thersholds back to original values, while leaving feature definitions alone
```
cademack: typo on thresholds here

howso/train.amlg
```diff
@@ -1,6 +1,20 @@
...
 
-	;records an entire session in one call, does not do any filtering.
+	;train the passed in cases, filtering out cases that match optionally passed in ablation parameters
+	; returns a response object in the following format:
+	;	(associate
+	;		"success" (true)/(false)
+	;		"num_trained" num_trained_cases
+	;		"ablated_indices" (list of ablated case indices)
+	;		"status" "status output message"
+	;		"message" "if failed to succeed, the error/reason message will be here"
```
cademack: it doesn't seem to me that "message" is a key in the returned assoc. Looks like if there is a "message", it gets stuffed in as an "error"

howso/upgrade.amlg
```diff
@@ -0,0 +1,395 @@
...
+	;instead of whatever they are encoded internally.
+	;as dict of:  { entity_id : { feature : value } }
+	;Decode nominal encoding, round and convert from epoch back to correct datetime formats as necessary
+	;
+	;parameters:
+	; trainee: name of trainee, reference for filename
+	; migration_filepath: path to migration folder where migration scripts are stored
+	#!export_cases_and_sessions_decoded
+	(seq
+
```
cademack: extra new line here

cademack (commented):
!lgtm

howso/get_cases.amlg
howsoRes: this is a GH issue. i'll fix whatever other whitespace issues are in the file while i'm at it though

howso/distances.amlg
howsoRes: GH issue.

performance_tests/asteroid_test.amlg
howsoRes: nah, no reason to add a method to return this code -  I'll just update the tests to use the same format as unit tests, so it'll be a 'howso' entity instead of the trainee code being directly in the test file

howso/hierarchy.amlg
```diff
@@ -298,11 +298,12 @@
 						)
 					)
 
-					;else load one flattened entity trainee and parse it out by 'call'ing its code to create all the contained entities
-					(let
-						(assoc temptrainee (call (load (concat filepath filename "." !file_extension))) )
-
-						(if (and (!= (null) temptrainee) (contains_entity temptrainee))
+					;else load entity as-is
+					(let (assoc temptrainee "_temptrainee")
```
howsohazard: Remove most of this logic -- load_entity returns null if the load failed, just use that.  Load into the name directly instead of moving.

cademack (dismissed):

cademack (approved):

PR #159, opened 2024-04-24
20009: Adds MCC as a prediction stat

jackx111:


howso.amlg
```diff
@@ -157,7 +157,7 @@
 	#!regionalModelMinSize (null)
 	#!regionalModelMinPercent (null)
 	#!revision 0
-	#!supportedPredictionStats (list "mae" "mda" "contribution" "confusion_matrix" "mda_permutation" "r2" "rmse" "spearman_coeff" "precision" "recall" "accuracy" "missing_value_accuracy")
+	#!supportedPredictionStats (list "mae" "mda" "contribution" "confusion_matrix" "mda_permutation" "r2" "rmse" "spearman_coeff" "precision" "recall" "accuracy" "missing_value_accuracy" "MCC")
```
howsohazard: Should be lower case like the rest

howso/feature_residuals.amlg
```diff
@@ -567,9 +567,9 @@
 				)
 		))
 
-		;map of nominal feature -> tuple of [accuracy, precision, recall, confusion_matrix]
+		;map of nominal feature -> tuple of [accuracy, precision, recall, MCC, confusion_matrix]
 		(assign (assoc
-			precision_recall_map
+			nominal_performance_map
```
howsohazard: since the continuous performance data is broken out into separate maps, at this point, does it make more sense to handle nominals in the same way?  maybe this is a question for @howsoRes

howso/prediction_stats.amlg
```diff
@@ -148,7 +148,7 @@
 	;
 	;parameters:
 	; stats: list of strings, optional.  Allowed values are:
```
howsohazard, line 150: see capitalization comment above

howso/feature_residuals.amlg
```diff
@@ -695,10 +696,54 @@
 											)
 											(size confusion_matrix)
 										)
+									; Generalized multi-class formula for Matthews Correlation Coefficient (MCC) is located https://en.wikipedia.org/wiki/Phi_coefficient#Multiclass_case
+									MCC
```
howsoRes: name the variable lower case mcc for consistency please

howso/feature_residuals.amlg
```diff
@@ -695,10 +696,54 @@
...
+														confusion_matrix
+													)
+												))
+												;get a list containing the predicted counts of each class, vector p in the MCC formula
+												predicted_counts
+												(append (values
+													(map 
+														(lambda 
+															(apply "+" (values (map
+																(lambda (or (get (current_value) (current_index 1))) 0)
```
howsoRes: add a comment what (current_index 1) is in this case

howso/feature_residuals.amlg
```diff
@@ -695,10 +696,54 @@
...
 										)
+									; Generalized multi-class formula for Matthews Correlation Coefficient (MCC) is located https://en.wikipedia.org/wiki/Phi_coefficient#Multiclass_case
+									MCC
+										(let 
+											(assoc 
+												;get the total correctly predicted counts, variable c in the MCC formula
+												total_predicted_correct
+												(apply "+" (values
+													(map (lambda (or (get (current_value) (current_index))) ) confusion_matrix)
+												))
```
howsoRes: not sure if this is a gh display issue, but make sure all the 'values' for each of the variables is indented relative to the variable, eg.:

```
total_predicted_correct
  (apply "+" ..
    ...
  )
```

howso/feature_residuals.amlg
jackx111: I agree that these should be consistent, I'll wait for @howsoRes's opinion

howso/feature_residuals.amlg
```diff
@@ -603,153 +599,216 @@
...
-											)
-											predicted_classes_map
+		;accuracy is: num correct predictions / all predictions
+		(assign (assoc
+			accuracy_map
+				(map
+					(lambda
+						(let 
+							;grab the current action feature's confusion matrix
+							(assoc current_confusion_matrix (get confusion_matrix_map (current_index 1)))
```
howsoRes: since confusion_matrix_map is a map of feature -> confusion_matrix, go ahead and get rid of the 'current_' in front, so it's just 'confusion_matrix'. and do it for the others below as well.   if you feel that it needs something prepended, then perhaps feature_confusion_matrix is more clear.

jackx111: yup, makes sense, was a vestige from some previous variable names

howsoRes (approved):

PR #160, opened 2024-04-25
20088: Adds disallow_subtrainee_file_operations flag to initialize method, MINOR

howsoRes:
adds flag to initialize method, if set to true,  create_subtrainee, load_subtrainee and save_subtrainee filepath parameter will be ignored and the stored value will be used instead.

### DIFF
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -236,6 +236,7 @@
 	; trainee_id: optional, unique id for trainee
 	#create_subtrainee
 	(seq
+		(call !ValidateFilepathParameter)
 		;create contained trainee inside the !traineeContainer
 		(assign (assoc trainee_path (append !traineeContainer trainee) ))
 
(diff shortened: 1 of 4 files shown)


howso/hierarchy.amlg
```diff
@@ -939,4 +942,46 @@
...
+
+		(map
+			(lambda
+				;if the env variable HOWSO_ENGINE_LIMITED_FS is set to true (not case sensitive), ignore filepath parameter
+				(if (and
+						(= "HOWSO_ENGINE_LIMITED_FS" (first (current_value)))
+						(contains_value (last (current_value)) "[tT][rR][uU][eE]" )
+					)
+					(conclude
+						(assign (assoc ignore_filepath_parameter (true) ))
```
cademack: Is there any reason to not just directly assign over `filepath` here in the `(map)`? rather than ticking the flag and doing it just below?

cademack (dismissed):

cademack (approved):

howsohazard (changes requested):
Not needed -- this should be handled elsewhere.

howsoRes:
not going to implement in engine but in a different layer

PR #165, opened 2024-05-02
20103: Refactors response output, MAJOR 

howsoRes:


howso/hierarchy.amlg
```diff
@@ -582,7 +582,7 @@
 							case_indices case_indices
 							output_ids (true)
 						))
-						(list "palyoad" "cases")
+						(list 1 "palyoad" "cases")
```
cademack: typo here on "payload" which looks like it was already there... might as well fix now

cademack (changes requested):

cademack (dismissed):

howso.amlg
```diff
@@ -790,25 +790,32 @@
...
+	; error_code: optional, string defining the error code, e.g., 'validation'
+	; error_details: optional, assoc defining field validation errors
 	; warnings: optional, list of warning strings to output
 	; payload: any value or object
 	#!Return
 	(if errors
-		(assoc "status" "error"	"errors" (map (lambda (assoc "detail" (current_value 1))) errors) "warnings" (null)	"payload" payload)
+		(list 0
+			(append
+				(assoc "details" (map (lambda (current_value )) errors) )
```
fulpm: The error response should include a key `detail` that is a string or list of strings. Optimally, it should only be a list if there are more than one message.

howso.amlg
```diff
@@ -790,25 +790,32 @@
...
-	;	'status' : 'ok', # string 'ok' or 'error'
-	;	'errors' : [ {'detail': 'Some error message'} ], # list of error dicts
-	;	'warnings' : [ {'detail': 'Some warning message'} ], # list of warning dicts
-	;	'payload' : whatever # json response payload
-	; }
+	; success:
+	; [ 1, { 'payload' : payload, 'warnings' : warnings } ]
+	;
+	; error:
+	; [ 0, { 'details': error_descriptions, 'code': error_code, 'errors': field_validation_details } ]
```
fulpm: docstring also needs to be updated for `detail`

cademack (approved):

PR #168, opened 2024-05-06
20155: Adds tests to verify case weights are reflected in prediction stats and residuals

cademack:
Adds two tests with pathological datasets that verifies case weights affect the resulting residuals and other prediction stats.

### DIFF
--- a/unit_tests/ut_h_weighted_residuals_cont.amlg
+++ b/unit_tests/ut_h_weighted_residuals_cont.amlg
@@ -72,6 +72,7 @@
             )
     ))
 
+    (print "MAE's are as expected: ")
     (call assert_approximate (assoc
         exp expected_residuals
         obs (map (lambda (get (current_value) "mae")) residuals)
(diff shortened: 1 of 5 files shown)


unit_tests/ut_h_weighted_residuals_cont.amlg
```diff
@@ -0,0 +1,87 @@
...
+            (get
+                (call_entity "howso" "get_prediction_stats" (assoc
+                    stats (list "mae")
+                    robust (false)
+                ))
+                "payload"
+            )
+    ))
+
+    (call assert_approximate (assoc
```
howsoRes, line 76: add a print statement above the assert to visually show what's being asserted when the test passes (like how other unit tests do it)

unit_tests/ut_h_weighted_residuals_nom.amlg
```diff
@@ -0,0 +1,108 @@
...
+                (call_entity "howso" "get_prediction_stats" (assoc
+                    stats (list "mae" "accuracy" "recall")
+                    robust (false)
+                    weight_feature "case_weight"
+                ))
+                "payload"
+            )
+    ))
+
+    (call assert_approximate (assoc
```
howsoRes, line 90: same comment as the other unit test

unit_tests/ut_h_weighted_residuals_cont.amlg
cademack: ah, I knew something felt missing.

howsoRes (approved):

PR #171, opened 2024-05-07
20010: Adds local prediction stats to react, MINOR

jackx111:


howso/details_residuals.amlg
```diff
@@ -30,30 +30,38 @@
...
+								store_values store_values
+								;use the same hyperparameters that were used to make the original prediction
+								custom_hyperparam_map hyperparam_map
+							))
+					))
+
+					;can't compute residuals or prediction stats if model is too small
+					(assoc
+						"residual_map" (zip relevant_features 0)
+						"prediction_stats" (zip relevant_features 0)
```
cademack: prediction stats should be an (assoc) for each feature, right? So `(zip relevant_features 0)` might not make much sense.

Maybe `(zip relevant_features (assoc))` or `(zip relevant_features (null))` would make more sense?

howso/details_residuals.amlg
```diff
@@ -30,30 +30,38 @@
...
 					)
-					;can't compute residuals if model is too small
-					(zip relevant_features 0)
 				)
 		))
 
 		;append feature deviations to output
-		(accum (assoc output (assoc "feature_residuals" local_feature_residuals) ))
+		(accum (assoc output (assoc "feature_residuals" (get local_feature_metrics "residual_map"))))
```
cademack: we may hit this code now when "feature_residuals" is not a requested detail. I suggest we wrap this again with an `(if (get details "feature_residuals")...`

howso/details_residuals.amlg
```diff
@@ -30,30 +30,38 @@
...
+
+		;append prediction stats to output if all statistics are computed
+		(if compute_all_statistics
```
cademack: this condition is probably correct... but I think it would be more obvious what's going on if it was `(if (get details "prediction_stats")..`

although this seems technically a bit faster. Just a suggestion

unit_tests/ut_h_react_explain.amlg
```diff
@@ -439,6 +439,34 @@
...
+				))
+				"payload"
+			)
+	))
+
+	(print "React Lv4 local prediction stats returned: ")
+	(call assert_approximate (assoc
+		exp 0.89
+		obs (get result (list "prediction_stats" "accuracy" "E"))
+		percent 0.5
```
cademack: might be worth trying to use the `thresh` parameter to assert_approximate rather than `percent`. I think `thresh` lets you set just a constant value that the difference cannot be greater than to pass.

I also think it seems like a better fit for the assertion below.

cademack (changes requested):

howso/details_residuals.amlg
jackx111: so I figured we always return feature residuals if we ask for prediction stats, since it is already calculated?

jackx111: modified

howso/details.amlg
```diff
@@ -89,6 +89,15 @@
 	;						locally around the prediction. Uses only the context features of the reacted case to
 	;						determine that area. Relies on 'robust_residuals' flag.
 	;
+	;		 "prediction_stats" true or false. If true outputs feature prediction stats for all (context and action) 
+	;						features locally around the prediction. The stats returned are ("r2", "rmse", "spearman_coeff",
+	;						"precision", "recall", "accuracy", "mcc"). The "confusion_matrix" may also be returned by setting
+	;						'confusion_matrix' to true. Uses only the context features of the reacted case to determine that 
+	;						area. Relies on 'robust_residuals' flag.
+	;
+	;		 "confusion_matrix" true or false. If true, will automatically set 'prediction_stats' to true and return the
```
cademack: "the confusion matrix".. 

This makes it sound like there would only be one. There should be one for each nominal feature, right?

howso/details.amlg
```diff
@@ -426,9 +435,35 @@
 			(call !ComputeCaseMDA)
 		)
 
+		(if (and (!= 0 (size (get details "prediction_stats"))) (not (get details "get_prediction_stats")))
```
cademack: `get_prediction_stats` was scrapped as a detail I think?

howso/details.amlg
```diff
@@ -426,9 +435,35 @@
...
+			(accum (assoc details (assoc "prediction_stats" (true))))
+		)
+
 		;append feature errors and deviations
-		(if (get details "feature_residuals")
-			(call !ComputeReactFeatureResiduals)
+		(if (or (get details "prediction_stats") (get details "feature_residuals"))
+			(seq
+				(if (get details "prediction_stats")
+					(assign (assoc
```
cademack: are `compute_all_statistics` and `store_values` declared earlier?

We should verify this `(assign)` is working as expected... I'd hate for these local stats to get cached in the global map.

howso/details_residuals.amlg
```diff
@@ -30,30 +30,53 @@
...
-		;append feature deviations to output
-		(accum (assoc output (assoc "feature_residuals" local_feature_residuals) ))
+		;append feature deviations to output if selected
+		(if (get details "feature_residuals")
+			(accum (assoc output (assoc "feature_residuals" (get local_feature_metrics "residual_map"))))
+		)
+
+		;append prediction stats to output if selected
+		(if (get details "prediction_stats")
+			(accum (assoc output (assoc
```
cademack: this doesn't match the style guidelines here.. should be more like
```
(accum (assoc
      output
            (assoc
                    prediction_stats
                           (append
                               ...... everything else is formatted correctly
                            )
            )
))
```

cademack: I also certainly didn't format this right in regard to spacing, please don't try to copy-paste. You will only be disappointed

howso/details_residuals.amlg
```diff
@@ -30,30 +30,53 @@
...
+				(append 
+					(remove 
+						(get local_feature_metrics "prediction_stats")
+						;Keep the confusion_matrix if it is selected for
+						(if (get details "confusion_matrix")
+							(list ".hyperparam_path" ".robust")
+							(list ".hyperparam_path" ".robust" "confusion_matrix")
+						)
+					)
+					(assoc "mae" (get local_feature_metrics "residual_map"))
```
cademack: keys in (assoc)'s don't need double quotes. This should be fine, but something to know.

cademack (changes requested):

howso/details.amlg
jackx111: yes

howso/details.amlg
jackx111: yes, forgot to remove

cademack (dismissed):

cademack (approved):

howsoRes (approved):

PR #173, opened 2024-05-08
19949: Adds experimental data reduction endpoint, MINOR

jdbeel:
Adds a (still experimental) data reduction endpoint in `ablation.amlg`

### DIFF
--- a/howso/ablation.amlg
+++ b/howso/ablation.amlg
@@ -139,7 +139,7 @@
 		(call !CreateCaseWeights (assoc feature_name weight_feature) )
 	)
 
-	;reduce a model by removing cases which have an influence weight entropy that falls above
+	;reduce the trained data by removing cases which have an influence weight entropy that falls above
 	; a threshold
 	;
 	;parameters:
@@ -147,7 +147,7 @@
 	; distribute_weight_feature: optional, default '.case_weight'. name of feature whose values to use as case weights
 	; influence_weight_entropy_threshold: optional, default 0.6. quartile to use as the cutoff threshold for
 	;	influence weight entropy when deciding which cases to remove.
-	#reduce_model
+	#reduce_data
 	(declare
 		(assoc
 			context_features (list)
--- a/howso/ablation.amlg
+++ b/howso/ablation.amlg
@@ -146,39 +146,22 @@
 	; context_features: a list of context features to use when computing influence weight entropies
 	; distribute_weight_feature: optional, default '.case_weight'. name of feature whose values to use as case weights
 	; influence_weight_entropy_threshold: optional, default 0.6. quartile to use as the cutoff threshold for
-	;influence weight entropy when deciding which cases to remove.
+	;	influence weight entropy when deciding which cases to remove.
 	#reduce_model
 	(declare
 		(assoc
 			context_features (list)
-			distribute_weight_feature (null)
-			influence_weight_entropy_threshold (null)
+			distribute_weight_feature !autoAblationWeightFeature
+			influence_weight_entropy_threshold !autoAblationInfluenceWeightEntropyThreshold
 
 			;internal
 			max_influence_weight_entropy_to_keep .infinity
 		)
 
-		(assign (assoc
-			distribute_weight_feature
-				(if
-					(= (null) distribute_weight_feature)
-					!autoAblationWeightFeature
-					distribute_weight_feature
-				)
-			influence_weight_entropy_threshold
-				(if
-					(= (null) influence_weight_entropy_threshold)
-					!autoAblationInfluenceWeightEntropyThreshold
-					influence_weight_entropy_threshold
-				)
-		))
-
-		(if
-			(not !hasPopulatedCaseWeight)
+		(if (not !hasPopulatedCaseWeight)
 			(call !InitializeAutoAblation)
 		)
-		(if
-			(not !hasInfluenceWeightEntropies)
+		(if (not !hasInfluenceWeightEntropies)
 			(call !ComputeAndStoreInfluenceWeightEntropies)
 		)
 
@@ -194,7 +177,7 @@
 				))
 		))
 
-		(call_entity trainee "!RemoveCases" (assoc
+		(call !RemoveCases (assoc
 			cases
 				(call !GetCasesByCondition (assoc
 					condition
--- a/howso/ablation.amlg
+++ b/howso/ablation.amlg
@@ -144,9 +144,9 @@
 	;
 	;parameters:
 	; features: a list of features to use when computing influence weight entropies
-	; distribute_weight_feature: optional, default '.case_weight'. name of feature whose values to use as case weights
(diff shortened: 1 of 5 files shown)


howso/ablation.amlg
```diff
@@ -139,6 +139,77 @@
...
 	)
 
+	;reduce a model by removing cases which have an influence weight entropy that falls above
+	; a threshold
+	;
+	;parameters:
+	; context_features: a list of context features to use when computing influence weight entropies
+	; distribute_weight_feature: optional, default '.case_weight'. name of feature whose values to use as case weights
+	; influence_weight_entropy_threshold: optional, default 0.6. quartile to use as the cutoff threshold for
+	;influence weight entropy when deciding which cases to remove.
```
howsoRes: add a tab or two between the ; and the words to denote that this is a continuation of the previous line and not a new parameter.

howso/ablation.amlg
```diff
@@ -139,6 +139,77 @@
...
+			distribute_weight_feature (null)
+			influence_weight_entropy_threshold (null)
+
+			;internal
+			max_influence_weight_entropy_to_keep .infinity
+		)
+
+		(assign (assoc
+			distribute_weight_feature
+				(if
```
howsoRes: for all the  simple if statements in this method, please put the condition on the same line as the `(if` statement.

howso/ablation.amlg
```diff
@@ -139,6 +139,77 @@
...
+		(assoc
+			context_features (list)
+			distribute_weight_feature (null)
+			influence_weight_entropy_threshold (null)
+
+			;internal
+			max_influence_weight_entropy_to_keep .infinity
+		)
+
+		(assign (assoc
```
howsoRes: is there a reason for this explicit assign instead of defaulting both these values in the declare-assign block above to the built-ins?

howso/ablation.amlg
```diff
@@ -139,6 +139,77 @@
...
+					(query_exists !internalLabelInfluenceWeightEntropy)
+					(query_quantile
+						!internalLabelInfluenceWeightEntropy
+						influence_weight_entropy_threshold
+						distribute_weight_feature
+					)
+				))
+		))
+
+		(call_entity trainee "!RemoveCases" (assoc
```
howsoRes: is this a copy-paste mistake (where did this code come from)? why is this a call_entity on a trainee instead of a `(call !RemoveCases` ?

howso/ablation.amlg
```diff
@@ -139,6 +139,77 @@
...
+			cases
+				(call !GetCasesByCondition (assoc
+					condition
+						(associate
```
howsoRes, line 184: just to make sure this is intentional - you are explicitly creating a temporary local variable with the same name to shadow the model parameter just for the duration of this call?  If this is intentional, please add a comment to explain why.  Shadowing variables like that isn't good practice IMO.

jdbeel: I don't follow. I'm using `associate` so that `!internalLabelInfluenceWeightEntropy` gets evaluated since the condition `assoc` needs to be  a feature name -> a list of two values representing the range of cases that satisfy the condition. So it ends up being

`(assoc influence_weight_entropy (list 1.045 .infinity)` or something similar.

unit_tests/ut_h_ablate.amlg
```diff
@@ -177,14 +177,32 @@
...
+	(assign (assoc
+		pre_reduction_size
+			(size (get (call_entity "howso" "get_cases")
+				(list 1 "payload" "cases")
+			))
+	))
+	(call_entity "howso" "reduce_model" (assoc influence_weight_entropy_threshold 0.5))
+
+	(print "Model reduction reduces model size by the expected amount: ")
+	(call exit_if_failures (assoc msg "Model reduction does not reduce model size by the expected amount"))
```
howsoRes: this should go after the assert_same call below, to exit if that assert fails

howso/ablation.amlg
```diff
@@ -139,6 +139,60 @@
 		(call !CreateCaseWeights (assoc feature_name weight_feature) )
 	)
 
+	;reduce the trained data by removing cases which have an influence weight entropy that falls above
+	; a threshold
+	;
+	;parameters:
+	; context_features: a list of context features to use when computing influence weight entropies
```
howsohazard: just use features; when we don't have context and action separated, just features is more consistent

howso/ablation.amlg
```diff
@@ -139,6 +139,60 @@
...
+	; distribute_weight_feature: optional, default '.case_weight'. name of feature whose values to use as case weights
+	; influence_weight_entropy_threshold: optional, default 0.6. quartile to use as the cutoff threshold for
```
howsohazard: quantile, rather than quartile.  also, should be more clear and explicit about whether the data is kept above or below this

howso/ablation.amlg
```diff
@@ -139,6 +139,60 @@
...
+	;	influence weight entropy when deciding which cases to remove.
+	#reduce_data
+	(declare
```
howsohazard, line 151: it would be really nice if we had a parameter to specify how far to reduce to, and then iterate until it's reached (or, if it can't ablate further due to some influence weight entropy threshold elbow, then stop then)

howso/ablation.amlg
```diff
@@ -139,6 +139,60 @@
...
 
+	;reduce the trained data by removing cases which have an influence weight entropy that falls above
+	; a threshold
+	;
+	;parameters:
+	; context_features: a list of context features to use when computing influence weight entropies
+	; distribute_weight_feature: optional, default '.case_weight'. name of feature whose values to use as case weights
+	; influence_weight_entropy_threshold: optional, default 0.6. quartile to use as the cutoff threshold for
+	;	influence weight entropy when deciding which cases to remove.
+	#reduce_data
```
howsohazard, line 150: can the threshold for doing this be set by auto ablation?  or is there a separate forthcoming PBI for that?

jdbeel: Yes, currently we use the threshold and  weight feature that are set by the auto-ablation methods (or overridden here.)

howso/ablation.amlg
jdbeel: That's coming next as a separate piece of work.

howso/ablation.amlg
jdbeel: Updated the docstrings to make this clearer

howso/ablation.amlg
```diff
@@ -139,16 +139,70 @@
 		(call !CreateCaseWeights (assoc feature_name weight_feature) )
 	)
 
+	;reduce the trained data by removing cases which have an influence weight entropy that falls above
+	; a threshold
+	;
+	;parameters:
+	; features: a list of features to use when computing influence weight entropies
```
howsohazard: should features also default to all regular features?

howsohazard (approved):

PR #178, opened 2024-05-14
20286: Adds `action_condition` and `context_condition` parameters to `get_prediction_stats`, MAJOR

jackx111:
Adds action_condition and context_condition parameter, along with several other parameters that accompanies those parameters. This allows the user to specify the action set and the context set. 

Also internally modifies CalculateFeatureResiduals to accept similar parameters.

### DIFF
--- a/howso/get_cases.amlg
+++ b/howso/get_cases.amlg
@@ -306,17 +306,19 @@
 	;	If 'condition_filter_query' is not specified:
 	;		If null, will be set to k if precision is "similar" or no limit if precision is "exact".
 	#!GetCasesByCondition
-	(declare (assoc
-		condition (assoc)
-		condition_filter_query (list)
-		condition_session (null)
-		precision "exact"
-		num_cases (null)
+	(declare 
+		(assoc
+			condition (assoc)
+			condition_filter_query (list)
+			condition_session (null)
+			precision "exact"
+			num_cases (null)
 		)
 	
 		(if (!= (null) condition_session)
 			;get all the case ids in the order they were stored for the specified session
 			(retrieve_from_entity condition_session ".replay_steps")
+
 			(seq
 				;if 'condition_filter_query' is not provided, get the 'condition_filter_query'
 				(if (= (size condition_filter_query) 0)
@@ -354,12 +356,13 @@
 	; precision: flag, default is "exact". Whether to query for 'exact' matches; if set to 'similar' will query for similar values intead of 'exact'
 	; num_cases: optional, default is null. The number of "similar" cases to get if 'precision' is "similar" or no limit if precision is "exact".
 	#!GetQueryByCondition
-	(declare (assoc
-		condition (assoc)
-		condition_session (null)
-		precision "exact"
-		num_cases (null)
-	)
+	(declare 
+		(assoc
+			condition (assoc)
+			condition_session (null)
+			precision "exact"
+			num_cases (null)
+		)
 
 		;set num_cases appropriately if null
 		(if (= (null) num_cases)
@@ -527,7 +530,6 @@
 			)
 		)
 	)
-
 
 
 	;retrieve the top or bottom number of cases for a specified feature, sorted top to bottom for top, and bottom to top for bottom
--- a/howso/get_cases.amlg
+++ b/howso/get_cases.amlg
@@ -289,9 +289,9 @@
 	;return all cases that match the specified condition
 	;parameters:
 	; precision: flag, whether to query for 'exact' matches; if set to 'similar' will query for similar values intead of 'exact'
-	; condition_session: optional, if specified ignores condition and returns cases for the specified session id
-	; condition_filter_query: optional, list of query conditions
-	; condition: optional assoc of feature->value(s). If 'condition_filter_query' is also provided, then 'condition_filter_query' overrides this parameter
+	; condition_session: optional, if specified ignores condition and condition_filter_query and returns cases for the specified session id
+	; condition_filter_query: optional, list of query conditions. If specified ignores condition and returns cases for the specified fitery query
+	; condition: optional assoc of feature->value(s). Condition for the returned cases. If 'condition_filter_query' is also provided, then this parameter is ignored.
 	;	no value = must have feature
 	;	- for continuous or numeric ordinal features:
 	;	one value = must equal exactly the value or be close to it for fuzzy match
--- a/howso/get_cases.amlg
+++ b/howso/get_cases.amlg
@@ -290,7 +290,8 @@
 	;parameters:
 	; precision: flag, whether to query for 'exact' matches; if set to 'similar' will query for similar values intead of 'exact'
 	; condition_session: optional, if specified ignores condition and returns cases for the specified session id
-	; condition: optional assoc of feature->value(s). If 'condition' is not provided, then 'query_condition' must be provided
+	; condition_filter_query: optional, list of query conditions
+	; condition: optional assoc of feature->value(s). If 'condition_filter_query' is also provided, then 'condition_filter_query' overrides this parameter
 	;	no value = must have feature
 	;	- for continuous or numeric ordinal features:
 	;	one value = must equal exactly the value or be close to it for fuzzy match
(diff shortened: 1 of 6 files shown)


howso/residuals.amlg
```diff
@@ -502,6 +502,18 @@
 							)
 						)
 
+						;if holding out, filter out all case ids provided
+						(declare (assoc
+							filtering_list
+							(if holdout case_ids (list case_id) )
```
cademack: This line should be indented, and the else condition should probably give an empty `(list)`

howso/prediction_stats.amlg
```diff
@@ -37,6 +37,7 @@
 	;			two values = inclusive between
 	;   	- for nominal or string ordinal features:
 	;			n values = must match any of these values exactly
+	; holdout: flag, optional. If set to true, the cases specified in 'condition' will be held out entirely when making predictions for calculating stats.
```
cademack: should probably add something like, "only used when condition is specified" or something along those lines

howso/residuals.amlg
jackx111: the else condition is the (list case_id), i'll format to look better

cademack (dismissed):

howso/get_cases.amlg
```diff
@@ -286,26 +286,80 @@
...
-	;		If null, will be set to k if precision is "similar" or no limit if precision is "exact". default is null
+	; precision: flag, default is "exact". Whether to query for 'exact' matches when using the 'condition' parameter; if set to 'similar' will query for similar values intead of 'exact'
+	;	If 'condition_filter_query' is provided, then this parameter is ignored.
+	; num_cases: optional, limit on the number of cases to get when using the 'condition' parameter. Works with or without 'condition_filter_query'. default is null.
+	;	If 'condition_filter_query' is specified:
+	;		If null there will be no limit.
+	;	If 'condition_filter_query' is not specified:
+	;		If null, will be set to k if precision is "similar" or no limit if precision is "exact".
 	#!GetCasesByCondition
 	(declare (assoc
```
cademack: drop `(assoc` down to the next line

cademack: see #get_cases a few methods above

howso/get_cases.amlg
```diff
@@ -286,26 +286,80 @@
...
+		condition_filter_query (list)
+		condition_session (null)
+		precision "exact"
+		num_cases (null)
+		)
+	
+		(if (!= (null) condition_session)
+			;get all the case ids in the order they were stored for the specified session
+			(retrieve_from_entity condition_session ".replay_steps")
+			(seq
```
cademack, line 322: I'd prefer a newline between the `(retrieve_from_entity)` and the `(seq`. It's hard to tell that these are the if/else branches without it

howso/get_cases.amlg
```diff
@@ -327,174 +381,155 @@
...
 						)
+
+						(list)
 					)
 				)
 			)
 		)
 	)
 
+
```
cademack, line 534: check the amount of newlines here.. seems like too many

howso/prediction_stats.amlg
```diff
@@ -30,16 +30,34 @@
 	; robust_hyperparameters: flag, optional. if specified, will attempt to return stats that were computed using hyperpparameters with the
 	;						  specified robust or non-robust type.
 	; weight_feature: string, optional. if specified, will attempt to return stats that were computed using this weight_feature.
-	; condition: assoc of feature->value(s)
+	; action_condition: assoc of feature->value(s), optional. If specified, will condition the action set, which is the dataset for which the prediction stats are for. 
+	;		If both 'action_condition' and 'context_condition' are provided, then all of the action cases selected by the 'action_condition' 
+	;       will be excluded from the context set, which is the set being queried to make to make predictions for the action set, effectively holding them out.
+	;		If only 'action_condition_filter_query' is specified, then only the single predicted case will be left out.
```
cademack: This isn't a parameter here? not sure what this is about

howso/prediction_stats.amlg
```diff
@@ -30,16 +30,34 @@
...
+	; action_condition_precision: optional string,  default is 'exact', used only with 'action_condition' parameter, will find exact matches if 'exact' and similar cases if 'similar'.
+	; action_num_cases: optional, limit on the number of action cases used in calculating conditional prediction stats. Works with or without 'action_condition_filter_query'.
+	;		If 'action_condition_filter_query' is set:
+	;			If null, will be set to k if precision is "similar" or no limit if precision is "exact". default is null
+	;		If 'action_condition_filter_query' is not set:
+	;			If null, will be set to the Howso default limit of 2000. default is null
+	; context_condition: assoc of feature->value(s), optional. If specified, will condition the context set, which is the set being queried to make to make predictions for the action set.
+	;		If both 'action_condition' and 'context_condition' are provided, then all of the cases from the action set, which is the dataset for which the prediction stats are for,
+	;       will be excluded from the context set,  effectively holding them out. If only 'action_condition_filter_query' is specified, then only the single predicted case will be left out.
```
cademack: more references to this.. was this a former parameter name?

howso/residuals.amlg
```diff
@@ -750,11 +790,21 @@
...
 													(list)
 												)
+												;if both action and context filter queries are provided while 'case_ids' were not provided at the beginning of the function, 
+												;only query the cases specifed by the context filter query, effectively holding out the action set. Otherwise query everything.
+												(if (and 
+														(not provided_case_id_flag)
+														(!= (size action_condition_filter_query) 0)
+														(!= (size context_condition_filter_query) 0)
+													)
+													context_condition_filter_query
```
cademack, line 830: I don't understand why this should only go here when `action_condition_filter_query` is also populated? It seems that with these new parameters we should be able to see the accuracy across any case of the model (no action_condition) using only a subset of the cases (non-empty context_condition), but this seems to suggest that this would not be the case. Is this intentional?

howso/residuals.amlg
```diff
@@ -431,6 +431,21 @@
...
+		;if 'case_ids' are not set, then use 'action_condition_filter_query' to get the case ids of the action set, otherwise
+		;flag that 'case_ids' were provided so 'action_condition_filter_query' is ignored.
+		(if (= (size case_ids) 0)
+			(assign (assoc
+				case_ids
+					(call !GetCasesByCondition (assoc
+						condition_filter_query action_condition_filter_query
+						num_cases num_samples
+					))
+			))
```
cademack, line 443: Not sure if it needs to happen here, or already happens somewhere else. But a consideration for Robust Residuals is that the total number of cases needs to be fairly large. The reasoning being that we need multiple predicitions with/without each feature. Say 12 cases match the query, this is not enough to stably compute robust residuals, so we should sample from those 12 with replacement

howso/prediction_stats.amlg
jackx111: yes supposed just to be action_condition

howso/residuals.amlg
jackx111: Yea, this was intentional, didn't seem like it was useful to see the entire of the entire model using a subset of the model. Certainly debatable

howso/prediction_stats.amlg
```diff
@@ -381,16 +404,34 @@
...
+	;			two values = inclusive between
+	;   	- for nominal or string ordinal features:
+	;			n values = must match any of these values exactly
+	; action_condition_precision: optional string,  default is 'exact', used only with 'action_condition' parameter, will find exact matches if 'exact' and similar cases if 'similar'.
+	; action_num_cases: optional, limit on the number of action cases used in calculating conditional prediction stats. Works with or without 'action_condition_filter_query'.
+	;		If 'action_condition_filter_query' is set:
+	;			If null, will be set to k if precision is "similar" or no limit if precision is "exact". default is null
+	;		If 'action_condition_filter_query' is not set:
+	;			If null, will be set to the Howso default limit of 2000. default is null
+	; context_condition: assoc of feature->value(s), optional. If specified, will condition the context set, which is the set being queried to make to make predictions for the action set.
```
howsoRes: "... to make predictions **on** the action set."   instead of `for`?

howsoRes (approved):

PR #180, opened 2024-05-18
20285: Adds support for sparse deviation matrix in hyperparameters, MINOR

howsoRes:
Adds 'use_sdm' parameter to analyze(); when true, if model optimizes to use deviations, all nominal features will attempt to use a sparse deviation matrix (SDM) instead of a single deviation value for their hyperparameter deviations.  The SDM is computed from each nominal feature's confusion matrix.

### DIFF
--- a/howso/analysis.amlg
+++ b/howso/analysis.amlg
@@ -915,7 +915,7 @@
 		(assoc
 			deviation_value
 				(if feature_is_nominal
-					;if the deviation is an assoc containing a SDM, pull the expected deviation value from it
+					;if the deviation is an assoc containing a sparse deviation matrix, pull the expected deviation value from it
 					(if (~ (assoc) feature_deviation)
 						(get feature_deviation "expected_deviation")
 
--- a/howso/analysis.amlg
+++ b/howso/analysis.amlg
@@ -890,6 +890,7 @@
 										0
 
 										(call !ConvertDeviationToFeatureWeight (assoc
+											feature_is_nominal (contains_index !nominalsMap (current_index 1) )
 											feature_deviation (current_value 1)
 											p_value (get output_map "p")
 										))
@@ -906,6 +907,7 @@
 
 	;helper method to convert deviation value to feature weight given a deviation and p value
 	;parameters:
+	; feature_is_nominal: flag, set to true if feature is nominal
 	; feature_deviation: value or tuple as passed in from the hyperparameter map
 	; p_value: p value from the hyperarameter map
 	#!ConvertDeviationToFeatureWeight
@@ -913,8 +915,26 @@
 		(assoc
 			deviation_value
 				;deviation may be a tuple if it contains null uncertainties
-				(if (= (list) (get_type feature_deviation))
-					(first feature_deviation)
+				(if (~ (list) feature_deviation)
+					;if nominal, make sure to grab deviation and not sparse deviation matrix
+					(if feature_is_nominal
+
+						;if first value is a tuple that contains the SDM, grab the value instead of the SDM
+						(if (~ (list) (first feature_deviation))
+							(get (first feature_deviation) 1)
+
+							;if first value is the SDM, grab the value instead of the SDM
+							(~ (assoc) (first feature_deviation))
+							(get feature_deviation 1)
+
+							;else just return the value
+							(first feature_deviation)
+						)
+
+						;else continuous, grab first value which is the actual devitaion
+						(first feature_deviation)
+					)
+					;else it's a single value, use it as-is
 					feature_deviation
 				)
 		)
@@ -1071,6 +1091,7 @@
 									0
 
 									(call !ConvertDeviationToFeatureWeight (assoc
+										feature_is_nominal (contains_index !nominalsMap (current_index 1) )
 										feature_deviation (current_value 1)
 										p_value p_value
 									))
--- a/howso/analysis.amlg
+++ b/howso/analysis.amlg
@@ -914,27 +914,15 @@
 	(let
 		(assoc
 			deviation_value
-				;deviation may be a tuple if it contains null uncertainties
-				(if (~ (list) feature_deviation)
-					;if nominal, make sure to grab deviation and not sparse deviation matrix
-					(if feature_is_nominal
+				(if feature_is_nominal
+					;if the deviation is an assoc containing a SDM, pull the expected deviation value from it
+					(if (~ (assoc) feature_deviation)
+						(get feature_deviation "expected_deviation")
 
-						;if first value is a tuple that contains the SDM, grab the value instead of the SDM
-						(if (~ (list) (first feature_deviation))
-							(get (first feature_deviation) 1)
-
-							;if first value is the SDM, grab the value instead of the SDM
-							(~ (assoc) (first feature_deviation))
-							(get feature_deviation 1)
-
-							;else just return the value
-							(first feature_deviation)
-						)
-
-						;else continuous, grab first value which is the actual devitaion
-						(first feature_deviation)
+						feature_deviation
 					)
-					;else it's a single value, use it as-is
+
+					;else continuous
 					feature_deviation
 				)
 		)
--- a/howso/analysis.amlg
+++ b/howso/analysis.amlg
@@ -43,6 +43,7 @@
 	; use_case_weights: optional, default is null. when true will scale influence weights by each case's weight_feature weight. if use_case_weights isn't specified, it will
 	;				   be true if auto ablation is enabled and false otherwise
 	; weight_feature: optional, default '.case_weight'.  name of feature whose values to use as case weights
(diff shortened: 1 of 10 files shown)


howso/analysis.amlg
```diff
@@ -903,15 +907,34 @@
...
 	; p_value: p value from the hyperarameter map
 	#!ConvertDeviationToFeatureWeight
 	(let
 		(assoc
 			deviation_value
 				;deviation may be a tuple if it contains null uncertainties
-				(if (= (list) (get_type feature_deviation))
-					(first feature_deviation)
+				(if (~ (list) feature_deviation)
+					;if nominal, make sure to grab deviation and not sparse deviation matrix
```
howsohazard: But the default deviation outside the SDM only applies to rare values that are not contained in the SDM.  This doesn't seem correct to me.

howso/hyperparameters.amlg
```diff
@@ -913,4 +944,233 @@
 
 		(call !Return)
 	)
+
+	;Convert confusion matrix (assoc of assoc of counts) to sparse deviation matrix (assoc of assoc of deviation)
+	;parameters:
+	; confusion_matrix:  assoc of class -> assoc of class -> predicted count for each class.
+	; feature: name of nominal feature
+	; smallest_count_threshold: number of predictions a class should have for it to count. If the count is less than this value, it'll be ignored.
+	;    default value is 10
```
howsohazard: If the count is less than this value, it won't be ignored, it will be in the combined deviations for the given value.

howso/hyperparameters.amlg
```diff
@@ -913,4 +944,233 @@
...
+					))
+					confusion_matrix
+				)
+			;map of class -> accuracy of random guessing
+			rand_guess_accuracy_map (map (lambda (/ (current_value) total_predictions)) predictions_per_class_map)
+		))
+
+		;indistinguishable accuracy
+		(declare (assoc
+			guessing_accuracy_matrix
```
howsohazard: needs more descriptive comment as to what this is -- could this be broken into its own method for readability?  same with some of these other parts of this method.  it's really long.

howso/hyperparameters.amlg
```diff
@@ -913,4 +944,233 @@
...
 		(call !Return)
 	)
+
+	;Convert confusion matrix (assoc of assoc of counts) to sparse deviation matrix (assoc of assoc of deviation)
+	;parameters:
+	; confusion_matrix:  assoc of class -> assoc of class -> predicted count for each class.
+	; feature: name of nominal feature
+	; smallest_count_threshold: number of predictions a class should have for it to count. If the count is less than this value, it'll be ignored.
+	;    default value is 10
+	#!ConfusionMatrixToSDM
```
howsohazard, line 986: In general, this method needs more comments and maybe to be broken apart for me to follow.  It also wasn't obvious where the random matching probability comes into play as a floor for accuracy.

howsohazard (changes requested):

howso/analysis.amlg
```diff
@@ -903,15 +907,22 @@
...
 	; p_value: p value from the hyperarameter map
 	#!ConvertDeviationToFeatureWeight
 	(let
 		(assoc
 			deviation_value
-				;deviation may be a tuple if it contains null uncertainties
-				(if (= (list) (get_type feature_deviation))
-					(first feature_deviation)
+				(if feature_is_nominal
+					;if the deviation is an assoc containing a SDM, pull the expected deviation value from it
```
howsohazard: spell out sparse deviation matrix most of the time, unless using it a bunch, in which case spell it out the first time in that method as: sparse deviation matrix (SDM)

howso/hyperparameters.amlg
```diff
@@ -913,4 +945,258 @@
 
 		(call !Return)
 	)
+
+	;Convert confusion matrix (assoc of assoc of counts) to sparse deviation matrix (assoc of assoc of deviation)
```
howsohazard: should mention optional lists here, where they are, and what is in them

howso/hyperparameters.amlg
```diff
@@ -913,4 +945,258 @@
...
+	;parameters:
+	; confusion_matrix:  assoc of class -> assoc of class -> predicted count for each class.
+	; feature: name of nominal feature
+	; smallest_count_threshold: number of predictions a class should have for it to have an explicit deviation value.
```
howsohazard, line 984: this method sometimes returns a null shortly below if it doesn't have enough data, which confused me.  need to accurately describe what it returns

howso/hyperparameters.amlg
```diff
@@ -913,4 +945,258 @@
...
+							(current_value)
+						)
+					))
+					guessing_accuracy_matrix
+				)
+		))
+
+		;store default class deviations for any classes that are not in the sparse confusion matrix
+		(declare (assoc
+			default_class_deviation_map
```
howsohazard: I don't like calling this deviation anymore, since default implies it is an average over everything, versus it could be an average over everything (if there's insufficient support for what's not in the matrix) or it could be just the remaining classes.  Would recommend something like uncommon_class_deviation_map or something similar

howsohazard: leftover_class_deviation_map?

howso/hyperparameters.amlg
```diff
@@ -913,4 +945,258 @@
...
+							(current_value)
+							(append (current_value) (associate predicted_class 0))
+						)
+					))
+				))
+				confusion_matrix
+			)
+	))
+
+	;helper method for computing SDM that clamps probabilities for the predicted classes vs random guessing
```
howsohazard: "to be no worse than random guessing"

howso/hyperparameters.amlg
```diff
@@ -913,4 +965,393 @@
...
+				))
+		))
+
+		;output null if the total count is too small
+		(if (< total_predictions smallest_count_threshold)
+			(conclude (null))
+		)
+
+		(declare (assoc
+			smallest_prob (/ 1 (+ 2 total_predictions))
```
cademack: a comment about the intuitions here for smallest/max would be nice

howso/hyperparameters.amlg
```diff
@@ -913,4 +965,393 @@
...
+							predicted_class (current_index 1)
+							row_map (current_value 1)
+						)
+						;leave entire row out if predicted class doesn't have enough samples
+						(if
+							(or
+								(= (null) (get row_map predicted_class))
+								(< (get row_map predicted_class) smallest_count_threshold)
+							)
+							row_map
```
cademack, line 1165: if the second condition passes, we return the whole row, but the comment suggests we should leave the row out. Should this return null instead?

howso/hyperparameters.amlg
```diff
@@ -913,4 +965,393 @@
...
+		))
+
+		(declare (assoc
+			;map of class -> accuracy of random guessing
+			rand_guess_accuracy_map (map (lambda (/ (current_value) total_predictions)) predictions_per_class_map)
+			all_classes_map
+				(compute_on_contained_entities (list
+					(query_value_masses
+						feature
+						(null)
```
cademack: I think this needs to be a bit smarter and actually use the weight feature if there is one in-use from the analyze call. I figure everything else should work "automatically" in the case of case weights because the confusion matrix should result from weighted sampling.

cademack (dismissed):

cademack (dismissed):

howso/hyperparameters.amlg
```diff
@@ -28,6 +28,45 @@
...
 			ordinal_feature_deviations (null)
 			attribute_map (assoc)
+			confusion_matrix_map (null)
+		)
+
+		(if (size confusion_matrix_map)
+			(assign (assoc
+				feature_deviations
+					(map
+						;convert deviations for nominals that have a confusion matrix into a assoc with the SDM
```
howsohazard: "with the SDM" -- I don't understand that part of the comment

howsohazard: do you mean converts the confusion matrix into an SDM?  better to spell out sparse deviation matrix here too

howso/hyperparameters.amlg
```diff
@@ -913,4 +965,399 @@
 
 		(call !Return)
 	)
+
+	;Convert confusion matrix (assoc class -> (assoc of class -> counts)) to sparse deviation matrix (SDM):
+	;with the possible formats:
+	;  (assoc of class -> (assoc of class -> deviation))
+	;  (assoc of class -> (list of: (assoc of class -> deviation), default_class_deviation))
+	;where default_class_deviation is the deviation value to be used for any classes that didn't have any counts.
```
howsohazard: is default_class_deviation an outdated comment given unspecified_class_deviation comment line 2 below?

howsoRes: default_class_deviation  are per-class (can have one for each row), whereas the unspecified_class_deviation is for the entire sdm

howsohazard (approved):

PR #185, opened 2024-05-24
20392: Parallelizes case deletion

jdbeel:


jdbeel:
I think we should wait until after the upcoming release to merge this.

PR #187, opened 2024-05-27
20401: Adds the ability to select features for prediction stats, MINOR

jackx111:


howso/prediction_stats.amlg
```diff
@@ -484,7 +489,14 @@
 					mode (if (= robust_hyperparameters (true)) "robust" "full")
 					weight_feature (if (= weight_feature (null)) ".none" weight_feature)
 				))
-			features (if (!= action_feature (null)) (append context_features action_feature) context_features)
+			features 
+				(if (!= prediction_stats_features (null))
+					(if (!= action_feature (null))
+						(list (values (append prediction_stats_features action_feature) (true)))
```
howsoRes: doesn't the outer (list) call wrap the output of (values) in another list, making a list of lists? Is this intended?

unit_tests/ut_h_stats.amlg
```diff
@@ -348,8 +348,7 @@
 		result
 			(call_entity "howso" "get_prediction_stats" (assoc
 				"robust" (false)
-				"stats" (list "accuracy" "precision")
-				"holdout" (true)
+				"stats" (list "accuracy" "precision" "confusion_matrix")
```
howsoRes: did you intend to remove the 'holdout' parameter here? that was being tested here per the comment prior to the assert_approximate, if it's removed, don't we expect accuracy and precision scores to be different? if they are still the same, was the holdout not being tested correctly?

jackx111: The holdout parameter was a leftover parameter from how the original holdout was implemented that I forgot to remove from the test. It is now unused.

howso/prediction_stats.amlg
jackx111: yes, was not intended, fixed

howso/prediction_stats.amlg
```diff
@@ -484,7 +489,14 @@
 					mode (if (= robust_hyperparameters (true)) "robust" "full")
 					weight_feature (if (= weight_feature (null)) ".none" weight_feature)
 				))
-			features (if (!= action_feature (null)) (append context_features action_feature) context_features)
+			features
```
cademack: I figure the declaration of `context_features` above is what should be altered in this change. It seems that context features could still be incorrect in flows where `prediction_stats_features` is specified.

howso/prediction_stats.amlg
```diff
@@ -60,6 +60,7 @@
 	;		If null, will be set to k. default is null
 	; num_robust_influence_samples_per_case: optional, Specifies the number of robust samples to use for each case for robust contribution computations.
 	;				  Defaults to 300 + 2 * (number of features).
+	; prediction_stats_features: list of string, optional. List of features to use when calculating prediction stats. Should contain all action and context features desired.
```
cademack: I feel like I'd rather this parameter just be "features" or something like that. This feels slightly redundant and verbose

unit_tests/ut_h_stats.amlg
howsoRes: Then please remove all the other instances of 'holdout' in this test as well (there's at least one above as well)

jackx111: done

howso/prediction_stats.amlg
```diff
@@ -60,6 +60,8 @@
 	;		If null, will be set to k. default is null
 	; num_robust_influence_samples_per_case: optional, Specifies the number of robust samples to use for each case for robust contribution computations.
 	;				  Defaults to 300 + 2 * (number of features).
+	; features: list of string, optional. List of features to use when calculating prediction stats. Should contain all action and context features desired. If
```
cademack: when calculating *conditional* prediction stats

unit_tests/ut_h_stats.amlg
```diff
@@ -361,9 +359,77 @@
...
+	; Fruits with small sizes have small being the minority size outside of strawberry.
+	; 	4/8 should be correct (all strawberries)
+	;	3/8 should be incorrect (apple, banana, peach)
+	;	1/8 should be random (1 null)
+	; There are 2 extra cases due to resampling for nulls.
+	; Min of 21/27 correct and max of 24/27 correct
+	(print "Output prediction stats selected features correctly: ")
+	(call assert_approximate (assoc
+		obs (get result "size")
+		exp (assoc
```
cademack: drop (assoc down to next line indented please

howsoRes (dismissed):

cademack (dismissed):

howso/prediction_stats.amlg
```diff
@@ -463,6 +467,8 @@
 	; num_robust_influence_samples_per_case: optional integer. Specifies the number of robust samples to use for each case.
 	;				  Applicable only for computing feature contributions. When unspecified, defaults to 200 + 2 * number of features.
 	;				  Higher values will take longer but provide more stable results.
+	; stats_features: list of string, optional. List of features to use when calculating conditional prediction stats. Should contain all action and context features desired. If
```
howsohazard: maybe prediction_stats_features?  It's long, but stats_features makes me think of something else.

howsohazard (approved):

PR #192, opened 2024-05-30
20434: Skips computing residuals and stats for unique features

howsoRes:
for unique features:
1) set deviations to smallest correct probability
2) skip computing residuals and stats 
3) default output to be null for nominal stats

### DIFF
--- a/howso/residuals.amlg
+++ b/howso/residuals.amlg
@@ -394,12 +394,12 @@
 		(if (size !uniqueNominalsSet)
 			(seq
 				(assign (assoc
-					unique_features_for_populating_output (filter (lambda (contains_index !uniqueNominalsSet (current_value))) features)
+					unique_features_for_populating_output (indices (keep (zip features) (indices !uniqueNominalsSet)))
 				))
 
 				(if (size unique_features_for_populating_output)
 					(assign (assoc
-						features (filter (lambda (not (contains_index !uniqueNominalsSet (current_value)))) features)
+						features (indices (remove (zip features) (indices !uniqueNominalsSet)))
 					))
 
 					;else reset unique_features_for_populating_output back to null
--- a/howso/residuals.amlg
+++ b/howso/residuals.amlg
@@ -94,9 +94,11 @@
 			!cachedFeatureMinResidualMap
 				(map
 					(lambda
-						;unique features don't have a residual
+						;For unique features:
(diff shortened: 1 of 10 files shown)


howso/residuals.amlg
```diff
@@ -388,6 +389,25 @@
...
+		;unique_features_for_populating_output to update the output at the end since unique features don't have residuals
+		(if (size !uniqueNominalsSet)
+			(seq
+				(assign (assoc
+					unique_features_for_populating_output (filter (lambda (contains_index !uniqueNominalsSet (current_value))) features)
+				))
+
+				(if (size unique_features_for_populating_output)
+					(assign (assoc
+						features (filter (lambda (not (contains_index !uniqueNominalsSet (current_value)))) features)
```
cademack: would something like `(indices (remove (zip features) (indices uniqueNominalsSet) ))` be better here?

cademack (dismissed):

howso/residuals.amlg
howsoRes: no because (indices) is o(n), (zip) is o(n), and the 2nd (indices) is also o(n).  that's 3 o(n) operations. whereas i have a (filter) which is just one o(n) and a (contains_index) which is a lookup.

cademack (dismissed):

howso/residuals.amlg
howsohazard: Yeah, but if you have 1000 features and only one unique, this code will take 1000 iterations.  Whereas Cade's would take 1.  It's unlikely that most of your features in a data set will be unique.  Though (values features (true)) is better than zip features.

howso/residuals.amlg
```diff
@@ -94,9 +94,11 @@
 			!cachedFeatureMinResidualMap
 				(map
 					(lambda
-						;unique features don't have a residual
+						;For unique features:
+						;Using n+1, n+2, or n+.5 are all possible considerations of Laplacian smoothing to apply a Bayesian approach for
+						;estimating the probability of having no incorrect predictions. We chose .5 assuming the Jeffreys prior approach
 						(if (contains_index !uniqueNominalsSet (current_index))
-							0
+							(/ 1 (+ 0.5 num_training_cases))
```
howsohazard, line 101: I thought there was at least one other place in the code where we have a minimum uncertainty... this change should be made there too.

howsohazard (approved):

PR #194, opened 2024-06-05
20494: Accumulates data mass when training, removing, and editing cases and uses that to determine when to auto-analyze

jdbeel:
- Adds a Trainee-level attribute, `!dataMassChangeSinceLastAnalyze`, which tracks the data mass that has changed since the last time the Trainee was analyzed
- Changes in data mass is accumulated when cases are: trained, removed, or edited
- Changed `!autoAnalyzeThrehsold` to represent the change in data mass required to trigger the next analyze rather than the raw number of cases required to trigger the next analyze.
- Update Trainee editing unit tests to confirm that `!dataMassChangeSinceLastAnalyze` is being edited across all of the Trainee editing endpoints.

### DIFF
--- a/howso.amlg
+++ b/howso.amlg
@@ -54,6 +54,7 @@
 	#!loadModulesEndpoint (null)
 
 	;value-storing labels, initialized and commented below via the initialize label
+	#!dataMassChangeSinceLastAnalyze (null)
 	#!numericalPrecision (null)
 	#!convictionLowerThreshold  (null)
 	#!convictionUpperThreshold  (null)
@@ -658,6 +659,9 @@
 			;for marginal stats like min, max, mode, mean, count, etc..
 			; { '.none': { 'feat_name': {'min': 2 ... } ... } ... }
 			!featureMarginalStatsMap (assoc)
+
+			;cumulative count of the data mass changed since the last analyze was called.
+			!dataMassChangeSinceLastAnalyze 0.0
 		))
 	)
 
--- a/howso.amlg
+++ b/howso.amlg
@@ -54,7 +54,6 @@
 	#!loadModulesEndpoint (null)
 
 	;value-storing labels, initialized and commented below via the initialize label
-	#!dataMassChangeSinceLastAnalyze (null)
 	#!numericalPrecision (null)
 	#!convictionLowerThreshold  (null)
 	#!convictionUpperThreshold  (null)
@@ -117,6 +116,7 @@
 	#!autoAnalyzeThreshold (null)
 	#!autoAnalyzeGrowthFactorAmount (null)
 	#!autoAnalyzeLimitSize (null)
+	#!dataMassChangeSinceLastAnalyze (null)
 	#!savedAnalyzeParameterMap (null)
 	#!derivedFeaturesSet (null)
 	#!sourceToDerivedFeatureMap (null)
@@ -583,7 +583,7 @@
 			;when false the model should not be auto-analyzed
 			!autoAnalyzeEnabled (false)
 
-			;if specified, stores the threshold for the number of cases at which the model should be re-analyzed
+			;if specified, stores the threshold for the change in data mass since the last analyze at which the model should be re-analyzed
 			!autoAnalyzeThreshold 100
 
 			;if !autoAnalyzeEnabledis set, the factor by which to increase that threshold everytime the model grows to that threshold size
--- a/howso.amlg
+++ b/howso.amlg
@@ -660,7 +660,8 @@
 			; { '.none': { 'feat_name': {'min': 2 ... } ... } ... }
 			!featureMarginalStatsMap (assoc)
 
-			;cumulative count of the data mass changed since the last analyze was called.
+			;cumulative count of the data mass changed since analyze was last called. this is accumulated to by any operations
+			; which add, remove, or edit cases.
 			!dataMassChangeSinceLastAnalyze 0.0
 		))
 	)
--- a/howso/analysis.amlg
+++ b/howso/analysis.amlg
@@ -543,7 +543,11 @@
 			(assign_to_entities (assoc
 				!autoAnalyzeThreshold
 					;!autoAnalyzeThreshold is a delta from the previous threshold.
-					(- (* num_cases !autoAnalyzeGrowthFactorAmount) !autoAnalyzeThreshold)
+					(max
+						(- (* num_cases !autoAnalyzeGrowthFactorAmount) !autoAnalyzeThreshold)
+						100 ;if several case deletions cause the above formula to return < 0, default
+							; to the default !autoAnalyzeThreshold.
+					)
 			))
 		)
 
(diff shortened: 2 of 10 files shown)


jdbeel:
Avoid merging until after current pending release.

howso/analysis.amlg
```diff
@@ -538,10 +538,15 @@
 			)
 			;set the next auto analysis threshold
 			(assign_to_entities (assoc
-				!autoAnalyzeThreshold (* num_cases !autoAnalyzeGrowthFactorAmount)
+				!autoAnalyzeThreshold
```
howsoRes, line 544: what's the logic for this change?  this breaks non-ablation logic for large batch training. ie, you threshold is at 100 but the first batch trains 1000, which is already bigger than the new threshold, it'll analyze but then immediately set it to be analyzed again, which may not be necessary if the next batch only trains 10.

howso/remove_cases.amlg
```diff
@@ -210,6 +210,11 @@
 					distribute_weight_feature distribute_weight_feature
 				))
 			)
+			;accumulate data mass change equivalent to one feature being changed for each case that satisfies the condition.
+			(accum_to_entities (assoc
```
howsoRes: single line please

howso/train.amlg
```diff
@@ -525,21 +525,35 @@
...
-						session (get_value session)
-						session_training_index (+ trained_instance_count (current_index 1))
-					))
+					(seq
+						;accumulates the data mass to !dataMassChangeSinceLastAnalyze, which is either 1.0
+						; or the case weight, whichever exists.
+						(accum_to_entities (assoc
+							!dataMassChangeSinceLastAnalyze
+								(if
+									accumulate_weight_feature
```
howsoRes: put condition on the same line as 'if' please, same with other if accumulate_weight_feature statement code below as well.

howso/update_cases.amlg
```diff
@@ -483,6 +489,14 @@
 					))
 					closest_cases_map
 				)
+
+				;also add the weight accumulated to each case to !dataMassChangeSinceLastAnalyze to ensure that cases trained as
+				; only weights (whether through auto-ablation or otherwise) contribute to the progress towards the next auto-analyze,
+				; if enabled.
+				(accum_to_entities (assoc
+					!dataMassChangeSinceLastAnalyze
+					1 ;(/ total_weight (size closest_cases_map))
```
howsoRes: don't leave code in comments, also this entire accum_to_entities should now be a one-liner, same wit the one below as well.

howso/analysis.amlg
```diff
@@ -538,10 +538,15 @@
...
+					;!autoAnalyzeThreshold is a delta from the previous threshold.
```
howsoRes, line 545: Also this statement is incorrect, !autoAnalyzeThreshold is not a delta. if you are you intending to make it a delta there will need to be a lot more refactoring as it is used in multiple other places for non-ablation (ie time series auto-analyze) flows.

howso/analysis.amlg
jdbeel: The logic was an attempt to turn this into a delta. Clearly there's more work to be done on this front before that change can be switched over in its entirety, as your other comment alludes to. I've pushed a change that should alleviate the problem that you've highlighted.

It shouldn't break non-ablation logic as adding (i.e., training) cases adds case_weight or 1 data mass so for most use-cases it's basically unchanged.

howso/analysis.amlg
jdbeel: Can you point out the other places where this is used? I pushed a change to the updating of the threshold above which should work (at least better) than the previous version. That, combined with the changes I've made so far has all of the tests passing + the RL recipe tests passing as well

howso/train.amlg
```diff
@@ -525,21 +525,34 @@
...
-								;else use feature_values as-is
-								feature_values
-							)
-						session (get_value session)
-						session_training_index (+ trained_instance_count (current_index 1))
-					))
+					(seq
+						;accumulates the data mass to !dataMassChangeSinceLastAnalyze, which is either 1.0
+						; or the case weight, whichever exists.
+						(accum_to_entities (assoc
```
howsoRes: this accum_to_entities winds up being called (size input_cases) number of times. this would be much more performant if you made the following change:
a) have this lambda/seq statement return the value of the if statement here (the accumulate_weight_feature value or 1.0)
b) store this entire map into a list of trained_case_weights
c) then do a single (accum_to_entities ... (apply "+" trained_case_weights))

howso/update_cases.amlg
```diff
@@ -483,6 +489,11 @@
 					))
 					closest_cases_map
 				)
+
+				;also add the weight accumulated to each case to !dataMassChangeSinceLastAnalyze to ensure that cases trained as
+				; only weights (whether through auto-ablation or otherwise) contribute to the progress towards the next auto-analyze,
+				; if enabled.
+				(accum_to_entities (assoc !dataMassChangeSinceLastAnalyze 1))
```
howsoRes: should this accum to a model parameter inside a map statement that iterates over all the cases be moved outside of the map statement since the total amount is known ahead of time so it can be called once?

howso/train.amlg
jdbeel: Didn't do this exactly the way you suggested, but instead pulled it out of the !TrainCreateCases method entirely, to similar effect.

howsoRes (dismissed):

howso.amlg
```diff
@@ -658,6 +659,10 @@
 			;for marginal stats like min, max, mode, mean, count, etc..
 			; { '.none': { 'feat_name': {'min': 2 ... } ... } ... }
 			!featureMarginalStatsMap (assoc)
+
+			;cumulative count of the data mass changed since analyze was last called. this is accumulated to by any operations
+			; which add, remove, or edit cases.
+			!dataMassChangeSinceLastAnalyze 0.0
```
cademack, line 665: I'd prefer this is listed closer to the other auto_analyze attributes

cademack (commented):
The comment for !autoAnalyzeThreshold should be updated I think. The current comment in howso.amlg describes it as the number of cases when the next analyze will be called, but now it describes the change in data mass that requires an analyze as far as I understand.

howso/train.amlg
```diff
@@ -222,6 +222,23 @@
 		;iterate over the input cases and create them, only returning case ids for cases that were able to be created
 		(declare (assoc new_case_ids (call !TrainCreateCases) ))
 
+		;accumulates the data mass change to !dataMassChangeSinceLastAnalyze, which is either 1.0
+		; or the case weight, if it exists.
```
cademack, line 226:
```suggestion
		; or the case weight, if it exists.
		
		(declare (assoc
		    weight_feature_index (get (zip features (indices features)) accumulate_weight_feature)
		))
		(accum_to_entities (assoc
			!dataMassChangeSinceLastAnalyze
				(apply "+"
					(map
						(lambda
							(if accumulate_weight_feature
								(or (get (current_value 1) weight_feature_index) 1)
								1
							)
						)
						input_cases
					)
				)
		))
```

howsoRes (approved):

PR #195, opened 2024-06-11
20482: Adds parameter validation and type hints

cademack:
This PR includes a full set of changes including type-hints/specifications for each public label's parameters.

Information about how to format and design the type-hints are added in contributing.md

The TLDR of the process is as follows:

1. Type hints are defined in the code as comments above each parameter of public labels that include an (assoc) of typing information following the guidelines detailed in engine_typing.md
2. When the Trainee is initialized, `#!BuildParameterValidationMap` is run which parses all of these type hints into code and stores them into `#!ParameterValidationMap`. This method also checks the parsed type hints to ensure they have valid types/references, nothing will be stored if *any* are invalid.
3. Larger/common type hints can be defined in the `#!CustomTypes` attribute in howso/typing.md
4. When a label is called, the first step should be a call to `#!ValidateParameters` which uses the `#!ParameterValidationMap` to check each parameter to see if it matches the given typing information.

To prevent merging/builds with invalid typing information, there are a couple preventative measures:
- `deploy_howso.amlg` early exits if `#!parameterValidationMap` was not populated in initialization
- a unit test: `ut_h_type_hints.amlg` was added which will further verify that the map can be populated correctly (also lists which methods/parameters have invalid type hints )

Things to be done that are not currently in this branch/PR:
- integration of parameter descriptions into parameter assocs provided by `#get_api`, for future OpenAPI spec creation purposes / documentation?
- documentation/specification of return types


CONTRIBUTING.md
```diff
@@ -57,12 +57,117 @@ called from `ut_comprehensive_unit_test.amlg`. If you want to test an endpoint,
...
+
+
+### Typing Examples
+
+Train
+```
+#train
+(declare
+    (assoc
+        ;(assoc type "list" values (assoc type "list" values "any"))
```
howsoRes: i suggest updating all the documented hints in these examples to use the { } paradigm since that's what we'll be using in the actual code base.

howso/typing.amlg
```diff
@@ -0,0 +1,159 @@
+(null
+
+    #!CustomTypes
```
howsoRes: This should be lowercase #!customTypes since it's not a method that is going to be called, but a datastore/lookup.
also needs a comment at the top of the file to match other modules, and another explaining what this particular store holds.

unit_tests/ut_h_type_hints.amlg
```diff
@@ -0,0 +1,28 @@
...
+			(list "BuildParameterValidationMap")
+		)
+	))
+
+    (declare (assoc
+        invalid_type_hints (call_entity "howso" "BuildParameterValidationMap")
+    ))
+
+    (call assert_true (assoc
+        obs (= (true) invalid_type_hints)
```
howsoRes, line 22: you don't need the (= (true)) here because that's what the assert_true tests, just do `obs invalid_type_hints`

howso.amlg
```diff
@@ -169,6 +169,8 @@
 	#minor_version 0
 	#point_version 0
 
+	#!parameterValidationMap (null)
```
howsoRes: seems to be missing a fat comment

CONTRIBUTING.md
cademack: good call

howso/typing.amlg
cademack: will do

howso.amlg
cademack: ha, indeed

unit_tests/ut_h_type_hints.amlg
cademack: well.. the behavior I need to capture here is if `invalid_type_hints` is (true) or if it is a (list) of stuff.  Both give a PASS on `#assert_true`. Hence why I have the specific `(= (true) _)` here. I'll keep it this way for now, but it may be worth tightening the `#assert_true` logic to specifically check if the value is `(true)` rather than if it is a "truthy".

howso/input_validation.amlg
```diff
@@ -249,4 +249,390 @@
...
+										))
+										(assign (assoc
+											exp_type (get specification "type")
+										))
+									)
+								)
+
+								(assign (assoc
+									optional
+										(if (contains_index specification "optional")
```
howsoRes, line 424: this works the same as this if block and is a one-liner: `(or (get specification "optional") (true))`

howso/input_validation.amlg
```diff
@@ -249,4 +249,390 @@
...
+					typing_map (current_value 1)
+					method_name "CustomTypes"
+					parameter_name (current_index 1)
+				))
+			)
+			!customTypes
+		)
+
+		(if (= 0 (size invalid_type_hints))
+			(assign_to_entities (assoc
```
howsoRes: on one line please

howso/input_validation.amlg
```diff
@@ -249,4 +249,390 @@
...
+				(call !CheckTypingMap (assoc
+					typing_map (current_value 1)
+					method_name "CustomTypes"
+					parameter_name (current_index 1)
+				))
+			)
+			!customTypes
+		)
+
+		(if (= 0 (size invalid_type_hints))
```
howsoRes: can simplify to: `(if (size invalid_type_hints) invalid_type_hints (assign_to_entities ..) )`

howso/hyperparameters.amlg
```diff
@@ -816,15 +830,22 @@
...
+			;{type "number"}
 			analyze_threshold 100
+			;{type "number"}
 			analyze_growth_factor 7.389056
+			;{type "number"}
 			auto_analyze_limit_size 200000
+		)
+		(call !ValidateParameters)
 
+		(declare (assoc
```
howsoRes: alright, so this method sucks in that actually it allows all the parameters to analyze to be specified as optional. notice how they are all being set on lines ~892-910 below, and these were actually meant to be as part of the input except they explicitly needed default values instead of just being null.

I think you'll have to copy them all from analyze into the parameter block of this method as well and undo this separation for these two.

howso/input_validation.amlg
cademack: dont think so.. if "optional" in `specification` was false, this would still return `(true)`

howso/hyperparameters.amlg
cademack: done

howso/input_validation.amlg
howsoRes: good point.

howso/react.amlg
```diff
@@ -15,25 +15,52 @@
...
 			feature_bounds_map (assoc)
+			;{type "list" values "string"}
 			preserve_feature_values (list)
+			;{type "boolean"}
 			substitute_output (true)
+			;{type "boolean"}
 			input_is_substituted (false)
+			;{type "string" enum ["min" "max" "most_similar"]}
 			new_case_threshold "min"
+			;{type "number"}
```
howsoRes, line 74: missing generate_new_cases parameter in all reacts (react_series as well)

howso/react_series.amlg
```diff
@@ -43,35 +43,60 @@
 	#react_series
 	(declare
 		(assoc
+			;{type "list" values "string"}
```
howsoRes, line 46: missing some parameters such as 'exclude_novel_nominals_from_uniqueness_check'.  I suggest cross-referencing howso-engine-py's core.py calls to make sure whatever parameters are being passed in there are also specified here

howso/input_validation.amlg
```diff
@@ -249,4 +249,407 @@
...
+			undefined_parameters (indices (remove given_parameters (indices (get !parameterValidationMap label_name))))
+		))
+
+		(if (size undefined_parameters)
+			(conclude (conclude
+				(call !Return (assoc
+					errors
+						(list (concat
+							"The following given parameters are not supported for " label_name ": "
+							(apply "concat" (trunc (weave undefined_parameters ", ")))
```
fulpm, line 387: Will want to make this is more robust once the clients are updated. This will eventually need to produce a general `detail` error message and provide an `errors` object containing the structure defined in ticket #20103. Will need to map out what `codes` we want to use for each error type when we get there.

CONTRIBUTING.md
```diff
@@ -57,12 +57,117 @@ called from `ut_comprehensive_unit_test.amlg`. If you want to test an endpoint,
...
+
+- All parameters must have at least "type" or "ref" specified
+- Optionality will be automatically inferred by the default value. Nullable/Optional are the same here.
+
+
+### Typing (assoc) keys and their values
+
+| Key    | Values |
+| ------ | ------ |
+| type | If single type, the type as a string. If multiple types, a list of types as strings. The available types are "list", "assoc", "number", "string", "boolean", "any".
```
jdbeel: Can you line up the pipes with the header?

CONTRIBUTING.md
```diff
@@ -57,12 +57,117 @@ called from `ut_comprehensive_unit_test.amlg`. If you want to test an endpoint,
...
+| indices | Only applicable when `type` is "assoc". Value is an assoc of named indices to their expected types.
+| additional_indices | Only applicable when `type` is "assoc". Value should be  similar to `type` values but is specified for indices that were not named in `indices`. This should only be used if `indices` is used as well.
+| optional | The value of this should be a boolean, and this should only be specified for sub-types. This will be automatically added to the top level of type hints based on the default value of the parameter.
+
+
+### Typing Examples
+
+Train
+```
+#train
```
jdbeel: Maybe try using made-up method names instead of using existing methods so that I don't feel compelled to have them match

howso/distances.amlg
```diff
@@ -192,15 +192,24 @@
 	#pairwise_distances
 	(declare
 		(assoc
+			;{type "list" values "string"}
 			features (null)
+			;{type "string"}
 			action_feature (null)
+			;{type "list" values {type "list" values "any"}}
```
jdbeel: Do both the `list` type hints and `assoc` type hints implicitly resolve their values to `any` if they are not specified? If so, it might be better to drop `values "any"` from these type hints throughout the PR

howso/input_validation.amlg
cademack: yep, sounds good. Will do then

howso/typing.amlg
```diff
@@ -0,0 +1,206 @@
...
+                        k "number"
+                        p "number"
+                        dt "number"
+                        featureDeviations (assoc type "assoc" values "any" optional (true))
+                        featureDomainAttributes (assoc type "assoc" values "any")
+                        featureWeights (assoc type "assoc" values "number" optional (true))
+                        paramPath (assoc type "list" values "string")
+                        useDeviations "boolean"
+                        allFeatureResidualsCached "boolean"
+                        !featureOrdinalDeviations (assoc type "assoc" values "any")
```
jdbeel: Is this correct? Should the key have an exclamation point in it or is it intended to resolve the value of `!featureOrdinalDeviations`?

howso/distances.amlg
cademack: dropped

howso/typing.amlg
cademack: its unfortunately real for now, we should resolve/remove this later though

howsohazard (approved):

PR #197, opened 2024-06-12
20501: Sparsifies confusion matrix, added 'confusion_matrix_min_count' param to react_into_trainee and get_prediction_stats,  MAJOR

howsoRes:
sparsifies confusion matrix by accruing all counts that are not statistically significant into separate values. This changes the format output for 'confusion_matrix' when calling get_prediction_stats

### DIFF
--- a/howso/feature_residuals.amlg
+++ b/howso/feature_residuals.amlg
@@ -16,7 +16,7 @@
 			use_case_weights (false)
 			weight_feature ".case_weight"
 			custom_hyperparam_map (null)
-			smallest_count_threshold 10
+			confusion_matrix_min_count 10
 		)
 
 		;if not using case weights, change weight_feature to '.none'
@@ -98,7 +98,7 @@
 					weight_feature weight_feature
 					custom_hyperparam_map hyperparam_map
 					compute_all_statistics (true)
-					smallest_count_threshold smallest_count_threshold
+					confusion_matrix_min_count confusion_matrix_min_count
 				))
 		))
 
@@ -263,7 +263,7 @@
 	; custom_hyperparam_map: optional, hyperparameters to use for residuals computation
 	; compute_all_statistics: flag, optional. if set to true will compute other statistics (precision, recall, r^2, rmse, etc.) in addition to MAE
 	; store_values: flag, optional. if set to true will store the statistics in trainee-level caches, if set to false will return the set of values
-	; smallest_count_threshold: number, optional, default is 10. Applicable only to confusion matrices, the number of predictions a class should have
+	; confusion_matrix_min_count: number, optional, default is 10. Applicable only to confusion matrices, the number of predictions a class should have
 	;		(value of a cell in the matrix) for it to remain in the confusion matrix. If the count is less than this value, it will be accumulated
 	;		into a single value of all insignificant predictions for the class and removed from the confusion matrix.
 	#!CalculateFeatureResiduals
@@ -287,7 +287,7 @@
 			custom_hyperparam_map (null)
 			compute_all_statistics (false)
 			store_values (true)
-			smallest_count_threshold 10
+			confusion_matrix_min_count 10
 		)
 
 		(declare (assoc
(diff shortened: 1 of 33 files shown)


cademack (changes requested):
(Putting this so it is not merged on accident)

howso/feature_residuals.amlg
```diff
@@ -1053,4 +1073,89 @@
...
+			)
+
+			(call !ComputeLeftoverConfusionMatrix)
+
+			;total of all statistically insignificant correct predictions
+			(declare (assoc
+				;create an assoc of class -> row from leftover_confusion_matrix for any rows where the predicted value isn't significant
+				satistically_insignificant_rows_map
+					(filter
+						(lambda (or (get (current_value) (current_index))) )
```
cademack: an (or) with one argument? I don't understand what this is about, shouldn't the value be able to be evaluated as a boolean anyway?

howso/feature_residuals.amlg
```diff
@@ -1053,4 +1073,89 @@
...
+					))
+				;total of all statistically insignificant incorrect predictions (total row count - correct prediction count)
+				total_leftover_incorrect_predictions
+					(apply "+" (values
+						(map
+							(lambda
+								;subtract predicted count from total of all counts for the row
+								(-
+									(apply "+" (values (current_value)))
+									(get (current_value) (current_index))
```
cademack, line 1135: Will the zero for the class being predicted always be here? Or should this be wrapped with an `(or __ 0)`?

howso/feature_residuals.amlg
```diff
@@ -1053,4 +1073,89 @@
...
+								(filter
+									(lambda (>= (current_value) smallest_count_threshold) )
+									(current_value)
+								)
+							)
+							;reduce confusion matrix to only those rows where predicted classes had significant counts
+							(remove confusion_matrix (indices satistically_insignificant_rows_map))
+						)
+					)
+				"leftover_correct" (+ (or total_leftover_correct_predictions))
```
cademack, line 1164: More of these one-arg (or)'s, please enlighten me :)

howso/feature_residuals.amlg
```diff
@@ -862,6 +862,26 @@
...
 
+		;wrap each confusion matrix in an assoc
+		(assign (assoc
+			confusion_matrix_map
+				;sparsify matrix only if running on enough cases or full model
+				(if (and
+						(size confusion_matrix_map)
+						(> (size case_ids) 200)
+					)
+					(call !SparsifyConfusionMatrices (assoc
```
cademack, line 883: So this method returns the (assoc) with the other keys as well. If we want to return these to the users, I think we should document somewhere what they are/what they mean. (probably needed for get_prediction_stats and the react detail comments).

If we don't want to return those extra keys (leftover counts, other counts, etc), then I suppose there should be a (remove) somewhere below this

howso/feature_residuals.amlg
howsoRes: numerify null if null

howso/feature_residuals.amlg
howsoRes: good point, the 'or' isn't necessary here

howso/feature_residuals.amlg
howsoRes: well yes, we'll need to update the api anyway.

currently it returns those extra keys when there are leftovers only, should it return those keys all the time instead?

howso/feature_residuals.amlg
howsoRes: since satistically_insignificant_rows_map only contains values that have the predicted class, we don't need to wrap it

howso/hyperparameters.amlg
```diff
@@ -1179,6 +1181,14 @@
 		(assign (assoc
 			leftover_confusion_matrix (filter (lambda (size (current_value))) leftover_confusion_matrix )
 		))
+	)
+
+	;Helper method for computing SDM, filters out counts below threshold and stores as probabilities correct instead.
```
howsohazard: For just computing SDM?  Would be better to spell out or mention the method(s) that it helps out

howso/feature_residuals.amlg
```diff
@@ -866,6 +866,26 @@
...
+						smallest_count_threshold 10
+						confusion_matrix_map confusion_matrix_map
+					))
```
howsohazard: Should expose smallest_count_threshold.  also, what is the logic around 200 case ids?

howso/feature_residuals.amlg
```diff
@@ -261,6 +263,9 @@
 	; custom_hyperparam_map: optional, hyperparameters to use for residuals computation
 	; compute_all_statistics: flag, optional. if set to true will compute other statistics (precision, recall, r^2, rmse, etc.) in addition to MAE
 	; store_values: flag, optional. if set to true will store the statistics in trainee-level caches, if set to false will return the set of values
+	; smallest_count_threshold: number, optional, default is 10. Applicable only to confusion matrices, the number of predictions a class should have
```
howsohazard: should name it something appropriate then -- maybe confusion_matrix_min_count?

howsohazard (approved):

cademack (approved):

PR #198, opened 2024-06-12
20505: Updates AccumulateCaseInfluenceWeights to utilize case weight values if given

cademack:
If a weight feature value is passed into !AccumulateCaseInfluenceWeights, then the accumulated case weights should be multiplied by that value rather than 1 as they are accumulated to nearest neighbors, similar to what is done in !DistributeCaseInfluenceWeights.

### DIFF
--- a/howso/update_cases.amlg
+++ b/howso/update_cases.amlg
@@ -427,6 +427,11 @@
 				))
 			;default value of 1 for the accumulate_weight_feature
 			new_weight_label_and_value (zip_labels (list accumulate_weight_feature) (list 1))
+			weight_feature_index
+				;will retrieve the index of weight feature or (null)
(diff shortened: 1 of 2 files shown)


howso/update_cases.amlg
```diff
@@ -435,6 +440,12 @@
 					case_values (current_value 1)
 					closest_cases_map (assoc)
 					total_weight 0
+					case_weight
+						(if weight_feature_index
```
howsoRes: this will statement fails when weight_feature_index is a valid value of 0

cademack: ah you're so right. Will make the condition a bit more explicit

howsoRes (approved):

PR #205, opened 2024-06-19
20608: Fixes reduce data removing all cases occasionally

jdbeel:
Fixes `reduce_data` for datasets where most cases have an `influence_weight_entropy` of 0. Previously, if too many cases had an `influence_weight_entropy` of 0 when `reduce_data` was called the threshold percentile had a chance of being equal to 0 which would remove all of the cases in the model.

Also does parameter cleanup and moves the query that is shared by data reduction and auto-ablation into its own method, to ensure unity.

### DIFF
--- a/howso/ablation.amlg
+++ b/howso/ablation.amlg
@@ -117,11 +117,7 @@
 
 		(call !StoreCaseValues (assoc
 			case_values_map influence_weight_entropy_map
-			label_name
-				(if (= (null) label_name)
-					!internalLabelInfluenceWeightEntropy
-					label_name
-				)
+			label_name (or label_name !internalLabelInfluenceWeightEntropy)
 		))
 
 		(accum_to_entities (assoc !revision 1))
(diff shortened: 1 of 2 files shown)


howso/ablation.amlg
```diff
@@ -222,10 +243,7 @@
 
 		(assign (assoc
 			max_influence_weight_entropy_to_keep
-				(compute_on_contained_entities (list
-					(query_exists !internalLabelInfluenceWeightEntropy)
-					(query_quantile !internalLabelInfluenceWeightEntropy !autoAblationInfluenceWeightEntropyThreshold ".case_weight")
-				))
+				(call !MaxInfluenceWeightEntropyToKeepQuery)
```
cademack, line 246: Not specifying the weight feature name here feels a bit scary. Is the autoAblationWeightFeature name changeable by the user (via set_auto_ablation_params, I would assume if so)

cademack (approved):

howso/ablation.amlg
jdbeel: Yeah, it is.

PR #207, opened 2024-06-25
20609: Updates to replace `.nan` and `.nas` with `(null)`

cademack:
Incoming Amalgam updates unify all nulls into the single `(null)`. With the removal of .nan and .nas, this PR update the Engine codebase to replace these values with `(null)`. Additionally simplifying the logic where possible.

### DIFF
--- a/howso/attributes.amlg
+++ b/howso/attributes.amlg
@@ -744,15 +744,15 @@
 			value_feature_nulls_are_dependent (get !featureAttributes (list value_feature "null_is_dependent"))
 		))
 
-		;if the continuous feature's attribute is to treat null as dependent, add .nas as a valid value for this dependent nominal feature if it has nulls
+		;if the continuous feature's attribute is to treat null as dependent, add (null) as a valid value for this dependent nominal feature if it has nulls
 		(if value_feature_nulls_are_dependent
-			;if this feature has any nulls, add .nas as a unique nominal feature value for which to compute boundaries
+			;if this feature has any nulls, add (null) as a unique nominal feature value for which to compute boundaries
 			(if
 				(compute_on_contained_entities (list
 					(query_equals feature (null))
 					(query_count)
 				))
-				(accum (assoc feature_values .nas))
+				(accum (assoc feature_values (list (null)) ))
 			)
 		)
 
@@ -768,14 +768,10 @@
 							(append base_query_among_statements (list
 								(query_among feature
 									(list
-										;convert .nas to (null) for query_among usage
-										(if (= .nas (current_index 3))
-											(null)
-											;ensure that the value in the query_among is a number by wrapping it in a (+) if the nominal is not a string
-											(if (contains_index !numericNominalFeaturesMap feature)
-												(+ (current_index 3))
-												(current_index 3)
-											)
+										;ensure that the value in the query_among is a number by wrapping it in a (+) if the nominal is not a string
+										(if (contains_index !numericNominalFeaturesMap feature)
+											(+ (current_index 3))
+											(current_index 3)
 										)
 									)
 								)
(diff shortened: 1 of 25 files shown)


howso/attributes.amlg
```diff
@@ -768,14 +768,10 @@
...
-											(null)
-											;ensure that the value in the query_among is a number by wrapping it in a (+) if the nominal is not a string
-											(if (contains_index !numericNominalFeaturesMap feature)
-												(+ (current_index 3))
-												(current_index 3)
-											)
+										;ensure that the value in the query_among is a number by wrapping it in a (+) if the nominal is not a string
+										(if (contains_index !numericNominalFeaturesMap feature)
+											(+ (current_index 3))
+											(current_index 3)
```
howsohazard: is the idea that if the feature is numeric and the value is null, a (+ (null)) is still a null?  Might be worth calling that out explicitly in the comment.

howso/details.amlg
```diff
@@ -678,10 +678,7 @@
 
 						;indices of a map are all strings, so convert them to correct types if necessary
 						(assign "hypothetical_action_values_map" (list feature)
-							(if (= .nas output_value)
-								(null)
-
-								(contains_index !numericNominalFeaturesMap action_feature)
+							(if (contains_index !numericNominalFeaturesMap action_feature)
```
howsohazard, line 682: also worth calling out the behavior of nulls here too in comments

howso/attributes.amlg
howsoRes: no the idea is that if this feature is a numeric nominal, this explicitly numerifies it, as per the comment on line 771.

howso/impute.amlg
```diff
@@ -150,9 +150,9 @@
 											))
 										)
 								)
-								;if entropy is ever nan, set the value to 1 so that only the number of nulls matters
+								;if entropy is ever (null), set the value to 1 so that only the number of nulls matters
 								;TODO: remove this statement after implementing 2992
```
howsohazard: 2992 was completed -- is this still relevant?

howso/attributes.amlg
howsoRes: the question is, is it still possible for feature_value to be null?  if so, then yes, the comment should be updated that nulls will remain nulls regardless of the path.

howso/marginal.amlg
```diff
@@ -489,7 +469,7 @@
 								;true means there are nulls, false means no nulls, null means unknown whether there are nulls
 								"has_nulls" (> num_nulls 0)
 								"null_residual"
-									(if (!= .nan (- max min))
+									(if (!= (null) (- max min))
```
howsoRes, line 472: could optionally replace this `if` block with one liner `or` like so:  `(+ (or (- max min) 0))`. it's fine fine either way tho :-)

howso/marginal.amlg
```diff
@@ -790,23 +770,23 @@
 								))
 							)
 
-							;output assoc of marginal stats for the feature, replace .nan with null if needed
+							;output assoc of marginal stats for the feature
 							(assoc
```
howsoRes, line 774: not saying you need do this, but this is a fitting place to use a { } style for the assoc, imo

howso/marginal.amlg
```diff
@@ -790,23 +770,23 @@
...
-								min (if (!= .nan min) min)
-								max (if (!= .nan max) max)
-								mean (if (!= .nan mean) mean)
-								median (if (!= .nan median) median)
-								percentile_25 (if (!= .nan percentile_25) percentile_25)
-								percentile_75 (if (!= .nan percentile_75) percentile_75)
-								mode (if (!= .nan mode) mode)
+								min min
```
howsoRes, line 775: not saying you need do this, but this is a fitting place to use a { } style for this assoc, imo

howso/attributes.amlg
cademack: added a comment that (null)s will be unaffected

howso/impute.amlg
cademack: It could still be (null) here, specifically when the model has < 2 cases. But I'll remove the TODO

howso/marginal.amlg
cademack: I'm gonna keep it this way as it's readable

howso/marginal.amlg
cademack: Don't really want to do a one-off here in this PR

howsohazard (approved):

PR #210, opened 2024-06-27
20674: Alters default values for react_series's "initial_values" parameter to demonstrate its optionality

cademack:


howso/react_series.amlg
```diff
@@ -145,6 +145,10 @@
 
 		(call !ValidateDerivedActionFeaturesIsSubset)
 
+		(if (= (null))
```
howsohazard: There is something missing in

cademack: fixed, thanks

howsohazard (approved):

PR #212, opened 2024-06-27
20416: Adds guidelines to reduce_data, MINOR

jdbeel:


jdbeel:
Don't merge yet.

howso/ablation.amlg
```diff
@@ -109,6 +109,229 @@
...
+			prev_prediction_stats_map (assoc)
+			new_prediction_stats_map (assoc)
+
+			;internal
+			abs_threshold_result_map (null)
+			delta_threshold_result_map (null)
+			output (assoc)
+		)
+
+		(assign (assoc
```
cademack: I see 2 of these result_maps declared above, but not the third (the rel one). I think you could just not declare the two earlier and change this block to a (declare)

howso/ablation.amlg
```diff
@@ -109,6 +109,229 @@
...
+			(accum "output" ["abs"] abs_threshold_result_map)
+		)
+		(if (call !FlattenThresholdResultMap (assoc threshold_result_map delta_threshold_result_map))
+			(accum "output" ["delta"] delta_threshold_result_map)
+		)
+		(if (call !FlattenThresholdResultMap (assoc threshold_result_map rel_threshold_result_map))
+			(accum "output" ["rel"] rel_threshold_result_map)
+		)
+
+		output
```
cademack: Maybe rather than (accum)'s to the empty (assoc) and returning it at the end, we could just build it at the end with something like:

```
(append
    (if (call !Flatten... )
        {"abs" abs_thres...}
        {}
    )
    (if (call !Flatten... )
        {"delta" delta_thres...}
        {}
    )
    (if (call !Flatten... )
        {"rel" rel_thres...}
        {}
    )
)
```

might be a little uglier to some, but would save us some copies?

howso/ablation.amlg
```diff
@@ -124,65 +347,138 @@
...
+			(assign (assoc
+				prev_prediction_stats_map
+					(get (call !CalculateFeatureResiduals (assoc
+						weight_feature distribute_weight_feature
+						use_case_weights (true)
+						compute_all_statistics (true)
+						store_values (false)
+					)) "prediction_stats")
+			))
+		)
```
cademack, line 411: the store_values flag *should* be dead? I'd check and remove if so

run_performance_tests.amlg
```diff
@@ -31,6 +31,11 @@
 	(direct_assign_to_entities (assoc run_online_retail (load "online_retail_test.amlg")))
 	(call run_online_retail)
 
+	; https://epistasislab.github.io/pmlb/profile/adult.html
+	#run_adult (null)
+	(direct_assign_to_entities (assoc run_adult (load "adult_test.amlg")))
+	(call run_online_retail)
```
cademack: calling the wrong label here

cademack (changes requested):
only one thing that really *needs* to be fixed :)

cademack (dismissed):

howso/typing.amlg
```diff
@@ -1117,6 +1117,20 @@
...
+                type "assoc"
+                description
+                    (concat
+                        "Map of prediction stats to map of feature name to threshold value. Thresholds can be absolute, delta, or relative depending "
+                        "on which parameter is being used."
+                    )
+                values
+                    (assoc
+                        type "assoc"
+                        values (assoc type "number")
```
cademack: we updated the typing to not support "values" for assocs, gotta use 'additional_indices' now for unnamed indices

cademack (approved):

PR #214, opened 2024-07-03
20777: Adds impute_react flag to internal react methods to not ignore future data during time series imputation

howsoRes:


howso/react.amlg
```diff
@@ -953,10 +953,12 @@
 		;for all time series where the time feature is a context,
 		;filter out future data based on whether the time feature is universal
 		(if !tsTimeFeature
-			(let
-				(assoc context_map (zip context_features context_values) )
-				(if (contains_index context_map !tsTimeFeature)
-					(accum (assoc custom_extra_filtering_queries (call !ComputeTimeSeriesFilterQuery) ))
+			(if (not impute_react)
```
howsohazard, line 957: and instead of nested ifs?

howsohazard (approved):

howso/react.amlg
howsoRes: not needed for tight loop performance. this block won't execute for most (non-ts) datasets

PR #216, opened 2024-07-10
20592: Removes imputation from time series training

howsoRes:


howso/derive_utilities.amlg
```diff
@@ -14,89 +14,140 @@
...
 						)
+					))
+					derived_features
+				))
+		))
 
-						;grab the latest case values
-						(assign (assoc current_case_map (zip features (last series_data))))
+		;return true because derivation failed and series should be stopped
+		(if derivation_failed (conclude (true)))
```
howsohazard, line 49: return?

howso/react_series_utilities.amlg
```diff
@@ -1348,6 +1349,58 @@
...
 	(while do_uniqueness_check
+
+		(assign (assoc
+			max_order
+				(apply "max"
+					(map (lambda (get !featureAttributes [(current_value 1) "ts_order"])) action_features)
+				)
+		))
+
+		;rate and delta features must not synth nulls at the start of a series where they are nulls
```
howsohazard: should add another comment that nulls in these positions will be handled specially (and where)

howso/derive_utilities.amlg
```diff
@@ -14,89 +14,140 @@
...
+						;feature needs to be generated if it references a row that's not in the current series_data or is out of bounds
+						(if (not valid_derivation)
 							(seq
-								(assign (assoc derivation_failed (true)))
-								(conclude)
+								;valid_derivation of (null) means a time feature generated a value past the max bondary, stop the series
+								(if (= (null) valid_derivation)
+									(assign (assoc derivation_failed (true) ))
+								)
+								;else output feature name that failed to derive so its value can be generated below
```
cademack: the "else" in this comment seems a bit dubious as it's not on an else branch, but should always be output when it was not a valid derivation

howso/derive_utilities.amlg
```diff
@@ -14,89 +14,140 @@
...
+		)
+
+		;iterate over invalid_derivation_features and synthesise their values instead
+		(map
+			(lambda (let
+				(assoc feature (current_value 1) )
+
+				;grab the latest case values
+				(assign (assoc current_case_map (zip features (last series_data))))
+				(declare (assoc
```
cademack: any reason to not move these 3 new vars into the (assoc) that follows the let?

howso/train.amlg
```diff
@@ -227,6 +227,44 @@
 			)
 		)
 
+		;if dataset has features that are for derivation output only and are not being explicitly trained on
+		;then add them to the training data with nulls
+		(if (< (size features) (size !trainedFeatures))
```
cademack: While this check is fast, I think it needs to be more proper such as `(size (remove (zip features) !trainedFeatures) )`.

This happens before !trainedFeatures is updated below. So here `features` could be a list of completely new features of the same number or more as !trainedFeatures.

cademack (dismissed):

howso/train.amlg
howsoRes: good point

howso/react_series_utilities.amlg
```diff
@@ -1357,7 +1357,10 @@
 				)
 		))
 
-		;rate and delta features must not synth nulls at the start of a series where they are nulls
+		;rate and delta features must not synth nulls at the start of a series where they are always nulls due to
+		;not having previous data, allowing other features such as value lags and time lags to determine whether the
+		;resulting value will output a value or a null.  If these rates/deltas were to ramain null, the resulting value
```
howsohazard: ramain -> remain

howsohazard (approved):

PR #217, opened 2024-07-10
20782: Adds ability to check cache first for details/stats and removes the need to call `react_aggregate` before case residual details, MINOR

jackx111:
also fixes a bug with feature_residual action features


howso/react_aggregate.amlg
```diff
@@ -462,32 +469,198 @@
...
 
+		;If we are not always recalculating metrics metrics, then we need to check which metrics exist in cache by attempting to retrieve them
+		(if (not always_recalculate)
+			(seq
+				(if
+					(= conditioned_residuals (false))
+					(assign
+						(assoc output
+							(call !GetFeatureDetailsStats (assoc
+								prediction_stats (get details "prediction_stats")
```
howsohazard: you can probably do this a lot more efficiently and cleaner via (append (keep details ....) (assoc strict_residuals (true)))

howso/react_aggregate.amlg
```diff
@@ -462,32 +469,198 @@
...
+								feature_mda_permutation_robust (get details "feature_mda_permutation_robust")
+								confusion_matrix_min_count confusion_matrix_min_count
+								strict_residuals (true)
+							))
+						)
+					)
+				)
+
+				;for all of the metrics, if they are selected, see if they exists when we attempt to retrieve them.
+				(if (get details "prediction_stats")
```
howsohazard: can all this be done with less code with a map?

howso/react_aggregate.amlg
```diff
@@ -462,32 +469,198 @@
...
+									stat_name "feature_mda_permutation_robust"
+								)
+							))
+					)
+				)
+
+				;if any metric is requested, but not cached, we set `not_all_metrics_cached` to indicate that we have to metrics to calculate and retrieve
+				(if
+					(or
+						(and (not cached_prediction_stats) (get details "prediction_stats"))
```
howsohazard: (apply "or" (...something))

unit_tests/ut_h_stats.amlg
```diff
@@ -388,6 +388,68 @@
...
+						selected_prediction_stats (list "accuracy" "precision" "recall" "mae" "spearman_coeff" "mcc" "rmse")
+					)
+				))
+				(list 1 "payload")
+			)
+	))
+
+	;for react_aggregate, 'always_recalculate' is turned off by default
+	(print "Output the same stats when react_aggregate is called twice, thus retrieving the cached values")
+	; (call assert_same (assoc
```
cademack: surely these asserts should be getting from result and result2?

jackx111: yes, forgot to change back

howso/react_aggregate.amlg
```diff
@@ -5,6 +5,9 @@
     ;
     ;parameters:
     ; context_features: optional list of features to use as contexts for computations. default is all features if unspecified.
+    ; always_recalculate: optional, flag, if set to true, will not attempt to retrieve the selected stats from the cache and will always relcalculate the stats,
```
cademack, line 8: not technically optional if it has a default value that isn't (null)

cademack: please update the rest too. Default is false, in which case it will check the cache first.

howso/analysis.amlg
```diff
@@ -62,6 +63,8 @@
 			;{type "boolean"}
 			bypass_calculate_feature_weights (null)
 			;{type "boolean"}
+			bypass_calculate_prediction_stats (null)
```
cademack: maybe this should default to false? rather than be optional

cademack: Res also says we should rename to "bypass_CACHE_prediction_stats" (capitalized for emphasis on the diff)

howso/hyperparameters.amlg
```diff
@@ -225,48 +225,105 @@
...
-						hyperparameter_feature action_feature
-						weight_feature weight_feature
-						use_case_weights use_case_weights
-						custom_hyperparam_map baseline_hyperparameter_map
-					))
 
+					)
 				)
 			)
+			(call !CalculateAndStoreFeatureResiduals (assoc
```
howsoRes: add a comment here saying ;else compute and cache residuals and stats using most current hyperparameters

howso/react_aggregate.amlg
```diff
@@ -462,32 +470,97 @@
 			(assign (assoc conditioned_residuals (true) ))
 		)
 
+		;If we are not always recalculating metrics, then we need to check which metrics exist in cache by attempting to retrieve them
+		(if (and (not always_recalculate) (= conditioned_residuals (false)))
+			(seq
+				(assign
```
cademack: could you change this to (assign (assoc)) on the same line? then indent output and the code that it evaluates to?

howso/hyperparameters.amlg
```diff
@@ -269,6 +269,7 @@
 					)
 				)
 			)
+			;else always calculate and cache residuals and prediction stats
```
howsoRes: add "using the latest hyperparameters"

howso/react_aggregate.amlg
```diff
@@ -462,32 +470,97 @@
...
+								weight_feature weight_feature
+								confusion_matrix_min_count confusion_matrix_min_count
+							)
+						))
+					)
+				)
+
+				(accum (assoc
+					existing_cached_stats
+						(values (map
```
cademack, line 494: This (map) looks like it returns a list, so (values) may be unnecessary.

howso/react_aggregate.amlg
```diff
@@ -594,81 +673,92 @@
...
-						feature_mda_full (get details "feature_mda_full")
-						feature_mda_permutation_full (get details "feature_mda_permutation_full")
-						feature_mda_robust (get details "feature_mda_robust")
-						feature_mda_permutation_robust (get details "feature_mda_permutation_robust")
-						confusion_matrix_min_count confusion_matrix_min_count
-					))
+			;if `not_all_metrics_cached` is true, then metrics have been calculated, and thus, need to be retrieved
+			(if (or not_all_metrics_cached always_recalculate)
+				(assign
+					(assoc output
```
cademack: (assign  (assoc)) on one line please

howso/react_aggregate.amlg
```diff
@@ -677,41 +767,40 @@
...
-					context_condition_precision (if
-						(= (get details "action_condition_precision") (null) )
-						"exact"
-						(get details "action_condition_precision")
-					)
+					action_condition_precision
+						(if (= (get details "action_condition_precision") (null) )
+							"exact"
+							(get details "action_condition_precision")
+						)
```
cademack: `(or (get details "action_condition_precision") "exact")`

I believe this should work the same way, if so do this for the spot below as well

howso/react_aggregate.amlg
```diff
@@ -677,41 +767,40 @@
...
+					context_condition_precision
+						(if (= (get details "action_condition_precision") (null) )
+							"exact"
+							(get details "action_condition_precision")
+						)
 				)
 				(assign
```
cademack: (assign (assoc)) one line please. I see this isn't from this PR, but I would still appreciate it

howso/react_aggregate.amlg
```diff
@@ -760,4 +849,84 @@
...
+		(declare (assoc exists (true)))
+
+		;prediction stats validation depends on the selected prediction stats
+		(if (= stat_name "prediction_stats")
+			(if
+				;if mae is the only prediction stat only, make sure every feature has a mae value
+				(and
+					(= (size selected_prediction_stats) 1)
+					(= (first selected_prediction_stats) "mae")
+				)
```
cademack: `(= selected_prediction_stats ["mae"])`

howso/react_aggregate.amlg
```diff
@@ -760,4 +849,84 @@
...
+				(seq
+					(map
+						(lambda
+							(if
+								(not (get output ["payload" (current_value 1) "mae"]))
```
cademack: I'd prefer this condition on the same line as the (if)

howso/react_aggregate.amlg
```diff
@@ -760,4 +849,84 @@
...
+				)
+				; else if it includes other prediction stats, check to see if any exists. If any
+				;  exists, then it means that they all should exist. This is checked this way since
+				;  prediction stats varies by feature type (nominal vs continuous).
+				(seq
+					(assign (assoc
+						selected_prediction_stats
+							(filter
+								;exclude mae, as mae can be returned even if the other prediction stats are not calculated
+								(lambda (not (contains_value ["mae"] (current_value))))
```
cademack: `(not (contains_value ["mae"] (current_value)))` should be `(!= "mae" (current_value))`

howso/react_aggregate.amlg
```diff
@@ -760,4 +849,84 @@
...
+								(call !ProcessSelectedPredictionStats (assoc selected_prediction_stats selected_prediction_stats))
+							)
+						exists (false)
+					))
+					(seq
```
cademack: This (seq) is in another (seq). So it is unnecessary

howso/react_aggregate.amlg
```diff
@@ -760,4 +849,84 @@
...
+
+		;prediction stats validation depends on the selected prediction stats
+		(if (= stat_name "prediction_stats")
+			(if
+				;if mae is the only prediction stat only, make sure every feature has a mae value
+				(and
+					(= (size selected_prediction_stats) 1)
+					(= (first selected_prediction_stats) "mae")
+				)
+				(seq
```
cademack: This pattern of
`(seq (map (lambda ... (assign (assoc exists (false))) )  features) exists)`
is a bit complicated. 

I suggest instead just doing a map that retrieves it all (dropping the `(not)`) and wrapping the map in a `(contains_value (map ...) (null))`. That way we get a boolean value which will be (true) if the stat is missing for any feature. Which we can wrap in a (not) to be more like what youre aiming at.

cademack (changes requested):

howso/react_aggregate.amlg
jackx111: actually changed the default to null, which it originally was

howso/react_aggregate.amlg
jackx111: looks like it becomes a list of list if i dont use values

PR #219, opened 2024-07-12
20890: Removes need to call react_aggregate before requesting case feature residual convictions in react

jackx111:
some of the test numbers slightly change because we no longer use sample_model_fraction in the test,

also includes a small bug fix for misnamed action feature parameter

### DIFF
--- a/howso/details_residuals.amlg
+++ b/howso/details_residuals.amlg
@@ -496,11 +496,7 @@
 						feature_residuals_robust (if robust_residuals (true))
 					)
 					use_case_weights (!= ".none" (last hp_param_path))
-					weight_feature weight_feature
-						(if (!= ".none" (last hp_param_path))
-							(last hp_param_path)
-							".case_weight"
-						)
+					weight_feature (if (!= ".none" (last hp_param_path)) (last hp_param_path) ".case_weight" )
 				))
 				(assign (assoc global_feature_residuals (get !residualsMap param_path)))
 			)
--- a/howso/details_residuals.amlg
+++ b/howso/details_residuals.amlg
@@ -506,18 +506,16 @@
 									".case_weight"
 								)
 					)
-					(seq
-						(call react_aggregate (assoc
-							action_feature action_feature
-							details (assoc
-								feature_residuals_full (if (not robust_residuals) (true))
-								feature_residuals_robust (if robust_residuals (true))
-							)
-							use_case_weights use_case_weights
-							weight_feature weight_feature
-						))
-						(assign (assoc global_feature_residuals (get !residualsMap param_path)))
-					)
+					(call react_aggregate (assoc
+						action_feature action_feature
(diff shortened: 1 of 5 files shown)


howso/details_residuals.amlg
```diff
@@ -485,77 +485,80 @@
...
-							(if (!= ".targetless" (first hp_param_path))
-								(concat ", action_feature='" (first hp_param_path) "'.")
-								", action_feature='.targetless'."
+						weight_feature
+								(if (not (= ".none" (last hp_param_path)))
+									(last hp_param_path)
+									".case_weight"
+								)
+					)
+					(seq
```
cademack: This (seq) is within a (let), should be able to just put the stuff inside without the (seq)

cademack (dismissed):

howso/details_residuals.amlg
```diff
@@ -485,77 +485,78 @@
...
 							(if (= ".none" (last hp_param_path))
-								" without case weights.\n"
-								(concat " with case weights from '" (last hp_param_path) "'.\n")
-							)
-							"Please call 'react_aggregate()', with "
-							(if robust_residuals "feature_residuals_robust=true" "feature_residuals_full=True")
-							(if (!= ".none" (last hp_param_path))
-								(concat ", use_case_weights=true, weight_feature='" (last hp_param_path) "'")
-								""
+								(false)
```
howsoRes: weird to have if statement that explicitly outputs a false/true.   just make the value (!= ".none" ... ) and it should evaluate to false/true as needed

howso/details_residuals.amlg
```diff
@@ -485,77 +485,78 @@
...
-							(if (!= ".targetless" (first hp_param_path))
-								(concat ", action_feature='" (first hp_param_path) "'.")
-								", action_feature='.targetless'."
-							)
-						))
-				))
-
-				(assign (assoc global_convictions_map (null) ))
+						weight_feature
+								(if (not (= ".none" (last hp_param_path)))
```
howsoRes: instead of  (not (=, why not a single != instead?

howso/details_residuals.amlg
```diff
@@ -485,77 +485,78 @@
...
-				(accum (assoc
-					warnings
-						(associate (concat
-							"Cannot compute 'global_case_feature_residual_convictions_full' because Full"
-							" global residuals have not been computed for action_feature "
-							(if (size action_features)
-								(concat "'" (first hp_param_path) "'")
-								"'.targetless'"
+				(let
+					(assoc
```
howsoRes: I suggest getting rid of this entire internal let-assoc and inline these values directly into the react_aggregate call as parameters since a) these variables aren't going to be reused anywhere else b) they are all simple one-liners

howsoRes (dismissed):

howsoRes (approved):

PR #220, opened 2024-07-12
20785: Adds logic to automatically determine `use_case_weights` when unspecified

cademack:


howso/conviction.amlg
```diff
@@ -112,6 +109,27 @@
 			model_size (call !GetNumTrainingCases)
 		))
 
+		(if (= (null) use_case_weights)
```
howsoRes: this exact block of code is used all over, can you please make it into a function where it has the parameter `set_valid_weight_feature`,  when 'set_valid_weight_feature' is true, it'll also include that 4 line `(if use_case_weights` portion as well.  that way you can just call this one method in all the places, even those that don't set valid_weight_feature

cademack: done!

howsoRes (dismissed):

howso/react_utilities.amlg
```diff
@@ -656,4 +656,37 @@
 			(list)
 		)
 	)
+
+	;helper method that update the values of use_case_weights, weight_feature, and optionally valid_weight_feature
```
howsohazard: update(s)
also, does this method return anything?  or does it modify on the stack?  would be good to clarify in the opening sentence

howso/conviction.amlg
```diff
@@ -37,7 +37,7 @@
 			;{type ["boolean" "string"]}
 			influence_weight_entropy (null)
 			;{type "boolean"}
-			use_case_weights (false)
+			use_case_weights (null)
```
howsohazard, line 41: this is missing in the docstring above -- fix any doc strings for other methods that are missing it

howso/react_utilities.amlg
```diff
@@ -656,4 +656,37 @@
 			(list)
 		)
 	)
+
+	;helper method that updates the stack's values of use_case_weights, weight_feature, and optionally valid_weight_featur
```
howsohazard: valid_weight_featur[e]

howsohazard (approved):

PR #222, opened 2024-07-16
20624: Streamlines train logic and improves training with ablation with large batches (20775)

howsoRes:
Breaks up ablation training to train in batches large enough to auto analyze after the batch is trained.
Also determine ahead of time whether autoanalyze should be run during the train, prior to storing cases, that way if it isn't necessary it won't do the check for every case.


howso/ablation.amlg
```diff
@@ -240,10 +240,14 @@
...
 		(assign (assoc
-			max_influence_weight_entropy_to_keep
-				(call !MaxInfluenceWeightEntropyToKeepQuery)
+			max_influence_weight_entropy_to_keep (call !MaxInfluenceWeightEntropyToKeepQuery)
 		))
 
+		;if there is no influence weight entropy cached yet, do the caching here
+		(if (= (null) max_influence_weight_entropy_to_keep)
+			(call !ComputeAndStoreInfluenceWeightEntropies (assoc features features))
+		)
```
howsohazard: shouldn't this do an assign to max_influence_weight_entropy_to_keep after calling?

howso/custom_codes.amlg
```diff
@@ -88,11 +96,11 @@
 						)
 				))
 
-				;remove the case_id column from series_data
+				;remove the case_id and !internalLabelSessionTrainingIndex columns from series_data
 				(assign (assoc
 					series_data
 						(map
-							(lambda (trunc (current_value)))
+							(lambda (trunc (current_value) -2))
```
howsohazard, line 104: comment the -2

howso/influences.amlg
```diff
@@ -614,11 +614,28 @@
 													;else this is a continuous feature, add the difference to the total
 													(let
 														(assoc action_feature (get action_features (get action_feature_indices (current_index 1))))
+
 														;targetless analysis deals with one action feature at a time, set index to be 0
+														(if targetless
+															(assign (assoc index 0 ))
+														)
+
+														;if a continuous feature isn't edit distance and isn't a numerical value, skip over it
```
howsohazard, line 623: it should still put the value in, but then warn the user.  is that what happens?

howsohazard (approved):

PR #225, opened 2024-07-16
20941: Adds opcode depth limits to custom and derived features to ensure stack overflows cannot occur with user specified code

howsohazard:


howso.amlg
```diff
@@ -496,6 +496,9 @@
 
 			;a value used to limit the allocations of each usage of (call_sandboxed), 0 = no limit
 			!sandboxedMemoryLimit 100000
+			
+			;a value used to limit the depth of opcode operations of each usage of (call_sandboxed)
+			!sandboxedOpcodeDepthLimit 50
```
howsoRes, line 502: need to declare this label uptop where others are declared, around line 167 or so

howsoRes (approved):

PR #229, opened 2024-07-19
20953: Adds multithreading to training with ablation and new ablation parameter: batch_size and ablated_cases_distribution_batch_size, MAJOR

howsoRes:
also gets rid of 'auto_analyze_limit_size' parameter.

### DIFF
--- a/howso/ablation.amlg
+++ b/howso/ablation.amlg
@@ -92,6 +92,9 @@
 			label_name (or label_name !internalLabelInfluenceWeightEntropy)
 		))
 
+		;compute and cache the max influence weight entropy to keep
+		(call !MaxInfluenceWeightEntropyToKeepQuery)
+
 		(accum_to_entities (assoc !revision 1))
 		(assign_to_entities (assoc !hasInfluenceWeightEntropies (true)))
 	)
@@ -157,6 +160,7 @@
 						(call !MaxInfluenceWeightEntropyToKeepQuery (assoc
 							influence_weight_entropy_threshold influence_weight_entropy_threshold
 							weight_feature distribute_weight_feature
+							cache_value (false)
 						))
 					cases
 						(call !GetCasesByCondition (assoc
@@ -185,6 +189,12 @@
 			)
 		)
 
+		;cache the most recent value
+		(call !MaxInfluenceWeightEntropyToKeepQuery (assoc
+			influence_weight_entropy_threshold influence_weight_entropy_threshold
+			weight_feature distribute_weight_feature
+		))
+
 		(accum_to_entities (assoc !revision 1))
 		(call !Return)
 	)
@@ -194,22 +204,34 @@
 	; influence_weight_entropy_threshold: optional, default !autoAblationInfluenceWeightEntropyThreshold. cases with
 	; 	influence weight entropy above this quantile will be removed
 	; weight_feature: optional, default !autoAblationWeightFeature. name of feature whose values to use as case weights
+	; cache_value: flag, default true.  when true will cache the computed value
 	#!MaxInfluenceWeightEntropyToKeepQuery
 	(declare
 		(assoc
 			influence_weight_entropy_threshold !autoAblationInfluenceWeightEntropyThreshold
 			weight_feature !autoAblationWeightFeature
+			cache_value (true)
 		)
 
-		(compute_on_contained_entities (list
-			(query_exists !internalLabelInfluenceWeightEntropy)
-			(query_not_equals !internalLabelInfluenceWeightEntropy 0)
-			(query_quantile
-				!internalLabelInfluenceWeightEntropy
-				influence_weight_entropy_threshold
-				weight_feature
-			)
+		(declare (assoc
+			quantile_value
+				(compute_on_contained_entities (list
+					(query_exists !internalLabelInfluenceWeightEntropy)
+					(query_not_equals !internalLabelInfluenceWeightEntropy 0)
+					(query_quantile
+						!internalLabelInfluenceWeightEntropy
+						influence_weight_entropy_threshold
+						weight_feature
+					)
+				))
 		))
+
+		(if cache_value
+			(assign_to_entities (assoc !autoAblationMaxInfluenceWeightEntropy quantile_value))
+		)
+
+		;output the value
+		quantile_value
 	)
 
 	;determine whether a new case (one that is not in the Trainee) should be ablated (trained as weights) or kept.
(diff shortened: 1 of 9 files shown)


howso/ablation.amlg
```diff
@@ -194,22 +204,34 @@
...
 
-		(compute_on_contained_entities (list
-			(query_exists !internalLabelInfluenceWeightEntropy)
-			(query_not_equals !internalLabelInfluenceWeightEntropy 0)
-			(query_quantile
-				!internalLabelInfluenceWeightEntropy
-				influence_weight_entropy_threshold
-				weight_feature
-			)
+		(declare (assoc
```
cademack, line 214: I think we may want to wrap this (compute_on_contained_entities) in some logic to check if its (null) or something like that. I think I remember Jacob ran into an issue where this ran when there were no cases that had the influence weight entropy cached or something like that. 

Just ensure that this value is usable, even if the model doesn't have the data for it.

cademack: oh it looks like this might be covered after the method call.. though the Trainee attribute could get the sketchy value

howsoRes: it is checked if it's null in the calling method, if it's null then it's null. i don't think we need to change anything

howso/hyperparameters.amlg
```diff
@@ -672,7 +672,6 @@
 						"auto_analyze_enabled" !autoAnalyzeEnabled
 						"analyze_threshold" !autoAnalyzeThreshold
 						"analyze_growth_factor" !autoAnalyzeGrowthFactorAmount
-						"auto_analyze_limit_size" !autoAnalyzeLimitSize
```
cademack, line 675: I can't comment on the exact line, but should we change the `(if)` condition here to be `(if !autoAnalyzeEnabled ...`?

howso/update_cases.amlg
```diff
@@ -440,84 +440,112 @@
...
-				)
-			))
-			cases
+						(current_value) ;closest_cases_map
+					)
+				))
+				closest_cases_maps
+			)
+
+			(accum (assoc case_index batch_size))
```
cademack: on the accum below to !dataMassChanceSinceLastAnalyze, it uses `(size cases)`. I believe we need to sum the case weights if those values are specified in the case data. basically if `weight_feautre_index` is not `(null)`

howso/ablation.amlg
```diff
@@ -194,22 +204,34 @@
 	; influence_weight_entropy_threshold: optional, default !autoAblationInfluenceWeightEntropyThreshold. cases with
 	; 	influence weight entropy above this quantile will be removed
 	; weight_feature: optional, default !autoAblationWeightFeature. name of feature whose values to use as case weights
+	; cache_value: flag, default true.  when true will cache the computed value
 	#!MaxInfluenceWeightEntropyToKeepQuery
```
howsohazard: Rename to something like RecomputeAndCacheMaxInfluenceWeightEntropy

howsohazard: That way you wouldn't need to have the comments everywhere stating that it's caching the value

howso/update_cases.amlg
```diff
@@ -440,84 +440,112 @@
...
-					case_weight
-						(if (!= (null) weight_feature_index)
-							(get (current_value 1) weight_feature_index)
-							;assume weight of 1 if weight feature values are not given
-							1
-						)
-				)
+		;accumulate all cases in batches of 100
+		(while (< case_index num_cases)
+			(assign (assoc batch_size (min 100 (- num_cases case_index)) ))
```
howsohazard: feels like this param of 100 should be exposed up quite a bit further, maybe into ablation params

howso/update_cases.amlg
```diff
@@ -440,84 +440,112 @@
...
-							(map 1 (filter (lambda (= (current_value) .infinity)) closest_cases_map) )
-					))
-				)
+						batch_of_cases
+					)
+			))
 
-				(assign (assoc total_weight (apply "+" (values closest_cases_map))))
+			;distribute the weights by writing to each of the local models' cases
+			(map
```
howsohazard: any performance to be gained here?  it probably spends too much time in entity lock bottlenecks, but I have to ask just in case

howso/hyperparameters.amlg
howsoRes: sure that's probably more readable

howso/update_cases.amlg
howsoRes: not sure - since this separate block is for writing to the entities, it's single threaded so it shouldn't have issues with entity locks.

howso/train.amlg
```diff
@@ -780,79 +763,123 @@
 
 	;Helper method to train cases with an ablation check
 	#!TrainCasesWitAblation
```
jdbeel: Shouldn't this be `#!TrainCasesWithAblation`?

howso/update_cases.amlg
```diff
@@ -440,90 +440,149 @@
...
-								;use a fixed random seed to guarantee deterministic behavior for reacts (named "fixed rand seed")
-								"fixed rand seed"
-								(null) ;radius
-								!numericalPrecision
+			;store all the local models for the cases in one pass
+			(assign (assoc
+				closest_cases_maps
+					||(map
+						(lambda (let
+							(assoc case_value_map (zip features (current_value 1)) )
```
jdbeel: It seems to me like `case_value_map` is only used on line 485 and is immediately unzipped. Couldn't we just drop the `(let` and just use `(current_value` directly in the query?

howso/train.amlg
```diff
@@ -780,79 +763,123 @@
...
+				)
+		))
 
-					(accum (assoc ablation_trained_instance_count 1))
+		;if auto analyzing, need to accumulate data masses for all cases
+		(if run_autoanalyze_check
+			(let
+				(assoc
+					mass_to_accumulate
+						(if (and accumulate_weight_feature (size indices_to_train))
```
cademack, line 849: this should probably be `(if (and weight_feature_index (size indices_to_train))`.

I don't see why it would be `accumulate_weight_feature`. Looks like it may have been incorrect before too if this was the case. Probably also worth verifying that `weight_feature_index` is available here as I think I saw that was moved.

howso/train.amlg
```diff
@@ -763,7 +744,9 @@
 						))
```
cademack, line 744: Could you fix the typo in `!TrainCasesWitAblation`

howso/hyperparameters.amlg
```diff
@@ -710,7 +708,6 @@
 			;{type "number"}
 			analyze_growth_factor 7.389056
 			;{type "number"}
```
cademack: should remove the type annotation comment as well

howso/hyperparameters.amlg
```diff
@@ -1012,6 +1004,10 @@
 			conviction_upper_threshold (null)
 			;{type "number"}
 			conviction_lower_threshold (null)
+			;{type "number"}
```
cademack: could you add `min 0` into the type annotations for both of these? :)

cademack: or maybe `min 1` makes more sense

howso/train.amlg
howsoRes: no, all the logic is based off accumulate_weight_feature, that is the boolean to do the accumulation and is true when there should be accumulation to whatever the weight_feature_index is, which could be 0, and thus can't be used to decide whether to do the accumulation.

howso/update_cases.amlg
```diff
@@ -440,90 +440,151 @@
...
+			batch_of_cases []
+			closest_cases_maps {}
+			entropy_weights []
+		))
+		(declare (assoc k_param (get hyperparam_map "k") ))
+		(declare (assoc
+			;cache the maximum entropy possibe given k similar cases, used to compute the entropy ratio for each accumulated case
+			max_influence_entropy (entropy (range (/ 1 k_param) 1 k_param 1))
+			;smallest possible amount to be distributed
+			smallest_entropy_amount (/ 1 (call !GetNumTrainingCases))
```
howsohazard: should be called smallest_case_weight or something similar, it's not entropy

cademack (dismissed):

howsohazard (dismissed):

howsohazard (approved):

PR #236, opened 2024-08-01
21066: Adds bounds check after post process in !ConvertToOutput

cademack:
Adds a bounding step after the values are modified according to the post processes.

### DIFF
--- a/howso/io.amlg
+++ b/howso/io.amlg
@@ -280,11 +280,11 @@
 
 												;if datetime, compare epoch to bounds and return max/min if appopriate
 												epoch_value
-												(if (> epoch_value (get !featureBoundsMap [(current_value) "epoch_max_value"]))
-													(get !featureBoundsMap [(current_value) "max"])
+												(if (> epoch_value (get !featureBoundsMap [(current_value 1) "epoch_max_value"]))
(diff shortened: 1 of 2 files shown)


howso/io.amlg
```diff
@@ -246,6 +246,59 @@
...
+					;bound according to feature bounds
+					(assign (assoc
+							decoded_values
+								(map
+									(lambda (let
+										(assoc
+											value (get decoded_values (current_index 1))
+											bounds_map (get !featureBoundsMap (current_value 1))
+											epoch_value
+												(if (contains_index !featureDateTimeMap (current_index 1))
```
howsoRes: (current_value 1) is the feature name, should be replaced everywhere where it should be feature name

howso/io.amlg
```diff
@@ -246,6 +246,59 @@
...
+													(format
+														(get decoded_values (current_index 1))
+														(get !featureDateTimeMap [(current_index 2) "date_time_format"])
```
howsoRes, line 262: also current_value

howsoRes (dismissed):

mkoistinen (commented):
There don't appear to be any unit tests to "protect" this functionality.

howsoRes (approved):

PR #238, opened 2024-08-05
20801:  Makes API changes in preparation for apiv3, MAJOR

jackx111:


howso/attributes.amlg
```diff
@@ -949,16 +949,16 @@
 	;example: (assoc "day" 7 "degrees" 360)
 	#!SetCyclicFeatures
 	(declare
-		(assoc features (assoc))
-		(if (= 0 (size features))
+		(assoc feature_attributes (assoc))
```
cademack, line 952: update the "docstring" above the method with the new parameter name

howso/details.amlg
```diff
@@ -71,7 +71,7 @@
 	;						model area. Uses only the context features of the reacted case to determine that area.
 	;
 	;		 "categorical_action_probabilities" true or false. If true outputs probabilities for each class for the action
-	;						features. Applicable only to categorical action features. react_series() calls additionally
+	;						features. Applicable only to categorical action features. single_react_series() calls additionally
```
cademack: can we drop the parenthesis here? these definitely don't make sense here in Amalgam land

howso/react_series_utilities.amlg
```diff
@@ -197,12 +197,12 @@
...
 	;
 	;parameters:
-	;	see comment block for #react_series in howso.amlg for details
+	;	see comment block for #single_react_series in howso.amlg for details
 	#!ReactSeries
 	(seq
 
 		(if (= 0 (size output_features))
-			;TODO: 15300 feed into react_group() and output results if details are provided
+			;TODO: 15300 feed into single_react_group() and output results if details are provided
```
cademack: can we remove the parenthesis here too

howso/react_utilities.amlg
```diff
@@ -42,7 +42,7 @@
 		)
 	)
 
-	;Generate lists of unique values for each 'unique' feature that isn't being conditioned on for use in #React
+	;Generate lists of unique values for each 'unique' feature that isn't being conditioned on for use in #single_react
```
cademack: should this be #SingleReact? seems to have changed from referencing the private label to the public label

migrations/migrations.amlg
```diff
@@ -271,5 +271,5 @@
 
 		(assign_to_entities updated_model_atributes_map)
 	)
-
-))
\ No newline at end of file
+"80.0.0"
```
cademack: what's this for? this makes it an invalid assoc

cademack: well invalid might be a strong word, it'll just have this key with a (null) value

jackx111: whoops, at a certain point i thought i needed this, was playing around

cademack (dismissed):

fulpm (dismissed):

cademack (approved):

PR #240, opened 2024-08-07
21118: Adds warnings to train and set_feature_attributes about violated feature bounds

cademack:
Adds warnings in two situations:
1. When new cases are trained that contain values which violate boundaries set in the feature attributes.
2. When feature attributes are set with boundaries that are violated by existing cases.

Also mildly refactors `#train` by moving a lot of the post-train logic into a new private helper method.


howso/attribute_maps.amlg
```diff
@@ -324,6 +324,62 @@
 				)
 			)
 
+			(if (> (call !GetNumTrainingCases) 0)
+				(seq
+					(if (contains_index bounds_map "min")
+						(if
+							(<
+								(retrieve_from_entity
+									(first (contained_entities (list
```
howsoRes: change this code and the one below to be:
```
(first (contained_entities
    (list (query_exists feature) (query_min feature 1))
))
```

PR #244, opened 2024-08-13
#21177 Fix performance-test.yml GitHub Action.

dmaze:
Since it's invoked from a separate workflow, it needs to be set up as `workflow_dispatch:`, and since it's not run as part of the core build, it needs to get the Howso Engine version and download it.

### DIFF
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -1,12 +1,21 @@
 name: Performance Test
 
 on:
-  workflow_call:
+  workflow_dispatch:
     inputs:
-      version:
-        required: true
-        type: string
       payload:
+        description: |
+          JSON payload containing extra info used by workflow.
+
+          For builds that have dependencies, key values of 'repo' => 'build-id' may
+          be used to define version. If blank or not supplied, latest tagged release
+          build will be used. Build ID may also be a release tag version.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/performance-test.yml
```diff
@@ -41,11 +58,19 @@ jobs:
...
+      run: |
+        gh ${{ needs.howso-engine-version.outputs.run-type }} download -D target -R "howsoai/amalgam" -p "*linux-amd64*" "${{ needs.howso-engine-version.outputs.run-id }}"
+        # Needed because release/non-release downloads are different structure
+        cd target && if [ ! -f howso-engine-*.tar.gz ]; then mv */howso-engine-*.tar.gz ./; fi && tar -xvzf howso-engine-*.tar.gz
+  
     - name: Test
       run: |
         sudo locale-gen es_ES.UTF-8
-        ./build.sh performance_test ${{ inputs.version }}
+        ./build.sh performance_test ${{ needs.howso-engine.outputs.run-id }}
```
apbassett: The parameter here needs to be the actual version of `howso-engine` and not the GitHub run ID, right? To get the version I think you'll need to parse `version.json` in the downloaded tarball.

apbassett (approved):

PR #246, opened 2024-08-13
#21177 Fix HSE repo name in perf test

dmaze:


.github/workflows/performance-test.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
       run: |
-        gh ${{ needs.howso-engine-version.outputs.run-type }} download -D target -R "howsoai/amalgam" -p "*linux-amd64*" "${{ needs.howso-engine-version.outputs.run-id }}"
+        gh ${{ needs.howso-engine-version.outputs.run-type }} download -D target -R "howsoai/howso-engine" -p "*linux-amd64*" "${{ needs.howso-engine-version.outputs.run-id }}"
```
apbassett:
```suggestion
        gh ${{ needs.howso-engine-version.outputs.run-type }} download -D target -R "howsoai/howso-engine" -p "howso-engine-*" "${{ needs.howso-engine-version.outputs.run-id }}"
```

apbassett: (The above wouldn't match the Engine build artifact, only Amalgam)

dmaze: That makes sense, except HSE only has one artifact so we don't need this `-p` option at all.  I've removed it.

apbassett (approved):

PR #251, opened 2024-08-14
21239: Reduces references in Trainee attributes

cademack:
Using (replace) to reduce the amount of references that are stored into the Trainee's attributes.

### DIFF
--- a/howso/feature_residuals.amlg
+++ b/howso/feature_residuals.amlg
@@ -620,7 +620,7 @@
 										;ensure to zip a list of classes by prepending an empty list
 										(append (list) (current_value))
 										;zip all classes with a value of 1 so that all clases in the predicted list are assigned a value of 1
-										1
+										(range (lambda (replace 1)) 1 (size (current_value)) 1)
 									)
 								)
 								(map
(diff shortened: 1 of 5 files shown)


howso/feature_residuals.amlg
```diff
@@ -620,7 +620,7 @@
 										;ensure to zip a list of classes by prepending an empty list
 										(append (list) (current_value))
 										;zip all classes with a value of 1 so that all clases in the predicted list are assigned a value of 1
-										1
+										(range (lambda (replace 1)) 1 (size (current_value)) 1)
```
jdbeel: The copy even makes a difference for this? A primitive declared in a lambda? The amount of these types of changes we're making would suggest there's language improvements to be made since this doesn't feel like a sustainable pattern to encourage...

jdbeel: At the very least, it's not a particularly intuitive/readable way to code.

jdbeel (dismissed):

howso/feature_residuals.amlg
howsohazard: To be clear, none of these changes that Cade has been implementing the past couple of days are required.  They're just to help with performance as these data structures are used in the tightest of loops.

howsohazard (approved):

PR #253, opened 2024-08-15
21180: Adds `action_features` parameter to react_aggregate, removes internal caching logic, and removes methods that accessed such caches, MAJOR

cademack:
Summary of changes:

`#react_aggregate` is very much rewritten to not use any of the Trainee level caches that were previously used for global statistics. All of these caches are removed as well as any public labels and private methods that were designed to retrieve from them. Global residuals used within the Engine are now instead stored within the hyperparameter map that was used to compute them under the key 'featureResiduals'. I also have changed all residuals-computing code to accept different lists for contexts/action features and brought null accuracy computations within the standard residual computing code.

API Changes in this PR:

- `action_features` is added as a parameter to `#react_aggregate`. This allows users to select the features that residuals/prediction stats will be computed for.
- the `global_case_feature_residual_conviction` React detail has been removed and the remaining `local_case_feature_residual_conviction` has been renamed to simply `case_feature_residual_conviction`.
- `#get_feature_contributions` has been removed.
- `#get_feature_residuals` has been removed.
- `#get_feature_mda` has been removed.


howso/hyperparameters.amlg
```diff
@@ -224,60 +224,14 @@
 			)
 		)
 
-		(declare (assoc deviations (get baseline_hyperparameter_map "!analyzeFeatureDeviations") ))
+		(if (= (first (get baseline_hyperparameter_map "paramPath")) ".default")
```
howsoRes: you can simplify this imo, since you're already doing a (get) opcode, may as well drop the (first) opcode and just provide the path ["paramPath" 0] instead

howso/details_stats.amlg
```diff
@@ -1,1057 +1,5 @@
...
-			(assign (assoc filtered_values_map original_values_map))
-
-			;else backup the filtered residuals map for further filtering below
-			(assign (assoc original_values_map filtered_values_map))
-		)
-
-		;at this point output the first one, whatever it happens to be
-		(first (values filtered_values_map))
-	)
-
```
howsoRes, line 1054: is the only method remaining in the file this !ProcessSelectedPredictionStats ? if so, a) is it still being used by anything and b) you should move it to wherever it's used get rid of this file altogether

howso/react_aggregate.amlg
```diff
@@ -372,360 +380,299 @@
...
+						))
+					selected_prediction_stats
+						(call !ProcessSelectedPredictionStats (assoc selected_prediction_stats (or (get details "selected_prediction_stats") [])))
+				)
+				(accum (assoc
+					output
+						(append
+							(filter
+								(lambda (contains_value selected_prediction_stats (current_index)))
+								(remove (get prediction_stats_map "prediction_stats") [".robust" ".hyperparam_path"])
```
howsoRes, line 444: can this ever result in a null? cos (append (null) {} {} ) will result in a badly formatted assoc

howso/hyperparameters.amlg
cademack: yep, good eye

howso/react_aggregate.amlg
cademack: Pretty sure it cannot result in a null, `(get prediction_stats_map "prediction_stats")` will always return an (assoc) when passing `compute_all_statistics (true)`

howso/details_stats.amlg
cademack: fair enough, I moved it.

howsoRes (approved):

PR #257, opened 2024-08-19
21189: Adds endpoint to enable to memory validation on !Return

cademack:
users can now call the `#enable_memory_validation` label to add a call to `(system "validate")` into `#!Return` which should help debug memory problems within the Engine.


apbassett (dismissed):

cademack:
We have decided this is not necessary and custom builds can be created for this purpose easily.

PR #264, opened 2024-08-26
20486: Implements shared deviations

jackx111:


howso/attribute_maps.amlg
```diff
@@ -195,6 +195,31 @@
...
+				;each group of shared deviations features is sorted
+				(assoc shared_features_list (sort (current_value 1)))
+				(map
+					(lambda (accum (assoc
+						shared_deviations_key_map
+							(append shared_deviations_key_map (associate (current_value 2) (first shared_features_list)))
+					)))
+					shared_features_list
+				)
+			))
```
cademack, line 349: please drop shared_deviations_map to the next line, just style reasons

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,31 @@
...
+			shared_deviations_features
+		)
+
```
cademack, line 351: is the (accum (assoc x (append x ...) ))
pattern here intentional? I'm struggling a bit to follow but accum-ing and append-ing to the same var at once seems strange

howso/feature_residuals.amlg
```diff
@@ -259,6 +259,7 @@
 			compute_all_statistics (false)
 			compute_null_uncertainties (true)
 			confusion_matrix_min_count 15
+			use_shared_deviations (true)
```
cademack: document flag please

howso/feature_residuals.amlg
```diff
@@ -1059,6 +1134,211 @@
...
+							(if (not (contains_value compressed_values (current_value)) )
+								(accum (assoc compressed_values (current_value 1)) )
+							)
+						)
+						(indices shared_deviations_map)
+					)
+					(map
+						(lambda
+							(if (= (null) (get compressed_values (current_index)) )
+								(if (!= (null) (get compressed_values (get shared_deviations_map (current_index)) ))
```
cademack: this block should be unindented

howso/feature_residuals.amlg
```diff
@@ -1059,6 +1134,211 @@
...
+									(accum (assoc
+										compressed_values
+											(associate
```
cademack: is this (get) missing a parameter?

howso/feature_residuals.amlg
```diff
@@ -1059,6 +1134,211 @@
...
+						(lambda
+							(if (not (contains_value compressed_values (current_value)) )
+								(accum (assoc compressed_values (current_value 1)) )
+							)
+						)
+						(indices shared_deviations_map)
+					)
+					(map
+						(lambda
+							(if (= (null) (get compressed_values (current_index)) )
```
cademack: I think I need a description/parameter descriptions here

howso/analysis.amlg
```diff
@@ -81,11 +81,21 @@
 			use_case_weights (null)
 			;{type "boolean"}
 			inverse_residuals_as_weights (null)
+			;{type "list" values "list"}
+			shared_deviations_features (null)
```
cademack: I don't know if this is just for testing or not... but I don't think this form of configuration is gonna fly. I was under the impression we were pretty hard-set on having this configured via feature attributes which would affect the model universally, not on a per-analysis basis. 

But maybe I missed a conversation

cademack (commented):
Left some comments on some nitpicky stuff and some other more meaningful stuff.

One thing I didn't comment on but I should mention overall - there are *a lot* of `(accum)` opcodes within loops here in these changes which is a pretty doomed pattern to use within Amalgam. We should definitely try to iron those out into something more functional if possible.

howso/feature_residuals.amlg
jackx111: no, it uses the value from the inner get

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,136 @@
...
+						;sort the list so duplicates can be removed
+						(sort (append [(current_value 1)] (get shared_deviations_map [(current_value 1) "shared_deviations"]) ))
+					)
+					(indices shared_deviations_map)
+				)
+			shared_deviations_key_map (assoc)
+		)
+
+		;get rid of duplicate groups
+		(assign (assoc shared_deviations_features_lists (values shared_deviations_features_lists (true))))
```
cademack: just wrap the above (map) with (values ... (true)), rather than assign onto the var again

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,136 @@
...
+
+		(map
```
cademack: try this instead:
```
(declare (assoc
		shared_deviations_key_map
			(apply "append" (map
				(lambda
					;(current_value) here is one of the inner lists ["A" "B" "C"]
					(zip (current_value) (first (current_value)))
				)
				shared_deviations_features_lists
			))
))
```

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,136 @@
...
+		)
+
+		;check to make sure that each feature only shows up once in any shared deviations groups
+		;flatten all of the values in the groups of shared deviations
+		(map
+			(lambda (accum (assoc
+				shared_deviations_features_flattened (current_value 1)
+			)))
+			(get shared_deviations_map "shared_deviations_features_lists")
+		)
```
cademack: check if this works:
```
(declare (assoc
    shared_deviations_features_flattened
        (apply "append" (get shared_deviations_map "shared_deviations_features_lists"))
))

File: howso/attribute_maps.amlg [hunk 11]
```diff
@@ -195,6 +195,136 @@
...
+						(=
+							current_shared_deviation_feature
+							(get shared_deviations_features_flattened (+ (current_index) 1))
+						)
+						(accum (assoc invalid_repeated_shared_deviations_features current_shared_deviation_feature ))
+					)
+				)
+			)
+			shared_deviations_features_flattened
+		)
```
cademack: This isn't our standard method of checking if something has duplicated values, we typically (zip) the list in question and see if the resulting (assoc) is the same size as the original list since keys cannot be duplicated.

However, idk if that method could be used to determine which keys are duplicated. I still think we could do something better than what is here though

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,136 @@
...
+
+		;check to see if any features in a shared deviations group is a unique nominal
+		(map
```
cademack: instead do a filter over the indices of the "shared_deviations_key_map" where the (lambda) is the your condition you are using within the (if). 

also instead of (contains_value (indices ...)), you could use (contains_index)

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,136 @@
...
+			)
+			(get shared_deviations_map "shared_deviations_key_map")
+		)
+
+		;warns if any of the errors are present
+		(if (or (size invalid_repeated_shared_deviations_features) (size invalid_unique_shared_deviations_features))
+			(seq
+				(if (size invalid_repeated_shared_deviations_features)
+					(accum (assoc
+						error_message
```
cademack: can we make `error_message` plural?

howso/feature_residuals.amlg
```diff
@@ -234,6 +234,8 @@
 	; confusion_matrix_min_count: number, optional, default is 10. Applicable only to confusion matrices, the number of predictions a class should have
 	;		(value of a cell in the matrix) for it to remain in the confusion matrix. If the count is less than this value, it will be accumulated
 	;		into a single value of all insignificant predictions for the class and removed from the confusion matrix.
+	; use_shared_deviations: flag, optional. If set to true, will return the residuals grouped by shared deviations group, otherwise will return the residuals
```
cademack: I feel like we'd be better off having this default to False and only turn it on within analyze or something

howso/residuals.amlg
```diff
@@ -118,73 +118,270 @@
...
+			)
+
+			(assign (assoc
+				temp_residual_map
+					(call !ExpandForSharedDeviations (assoc
+						compressed_values temp_residual_map
+					))
+			))
+
+			temp_residual_map
```
cademack: if we assign to a variable and then immediately return it, (as long as there are no other things being statefully changed), we can instead just return the value that we are assigning with immediately.

so instead of:
``` 
(assign (assoc temp_residual_map (call ....)  ))
temp_residual_map
``` 

you should probably just do
```
(call ...)
```

cademack (changes requested):

howso/attribute_maps.amlg
jackx111: switched to only checking one error message

howso/attribute_maps.amlg
jackx111: no longer doing this check

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,194 @@
...
+		)
+
+		;check to see if nominal features has a shared deviations with non-nominal features
+		(if (size nominals)
+			(declare (assoc
+				invalid_nominals_non_nominals_shared_deviations
+					(filter
+						(lambda
+							(let
+								(assoc
```
howsohazard: These variables would be better as two filters rather than a map with assigns in it.

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,194 @@
...
+									(lambda
+										(if (contains_value nominals (current_value))
+											(assign (assoc contains_nominals (true)) )
+											(assign (assoc contains_non_nominals (true)) )
+										)
+									)
+									;indices of an assoc of shared deviations groups, i.e. {"a" (null) "b" (null)} -> ["a", "b"]
+									(indices (current_value))
+								)
+								(and
```
howsohazard: Since these are boolean variables, it's overkill to check = true.  Also, if the filters from above are small enough, they can be put right here, and this whole block can be shrunk and sped up considerably!

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,133 @@
...
+	;  for calculations If an assoc, adds back in removed features as keys and copies the value of the primary key for each shared deviation group as its value.
+	;       e.g. {"a": 2.5, "b": 0} -> {"a": 2.5 "b":9 "c": 2.5 "d": 2.5} if features a, c, d were in a shared deviations group.
+	#!ExpandForSharedDeviations
+		(declare
+			(assoc
+				compressed_values (null)
+				shared_deviations_map !sharedDeviationsMap
+			)
+			(if shared_deviations_map
+				(if (~ (list) compressed_values)
```
howsohazard: this looks to be a lot of accums in maps, especially for something large and wide.  reduce may be a better opcode for this structure.

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,133 @@
...
+	; feature_residuals_lists: An assoc where the key is the variable name under which the prepped features residuals list will be stored after prep, the value is the feature residuals lists to be prepped.
+	#!PrepSharedDeviations
+	(declare
+		(assoc
+			features_list (assoc)
+			feature_residuals_lists (assoc)
+			expanded_mapping (assoc)
+		)
+
+		(map
```
howsohazard: See comment above -- reduce may be better for this as well.

howso/residuals.amlg
```diff
@@ -51,24 +51,72 @@
...
+								(lambda
+									(let
+										(assoc
+											current_feature_smallest_gap
+												(compute_on_contained_entities (list
+													(query_min_difference (current_value 2) (get !cyclicFeaturesMap (current_value 2)) )
+												))
+										)
+										(if
+											;if the current gap for the feature is smaller than the smallest feature gap in the group so far
```
howsohazard: This map may be able to be cut down to an (apply (min...

howso/residuals.amlg
```diff
@@ -129,62 +191,173 @@
...
+											(call !GetSharedDeviationGrouping (assoc feature_group_to_retrieve (current_index 2)))
+											(null)
+										)
+									class_counts_map (assoc)
+									current_feature (current_index 1)
+								)
+								(if (> (size feature_grouping) 0)
+									;if a feature with shared deviations, use the class counts from all the features
+									; in the group
+									(map
```
howsohazard: Also may be a good reduce candidate.

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,133 @@
...
+				)
+			)
+			(sort compressed_values)
+		)
+
+	;Helper function to prep shared feature deviations, by taking in a list of features and their list of residuals,
+	; combining all of the residuals under the primary feature key of the group of the shared deviations, and removing the features
+	; from the features list whose residuals have been combined into another features.
+	;parameters:
+	; features_list: An assoc where the key is the variable name under which the prepped features list will be stored after pre, the value is the feature list to be prepped.
```
howsohazard: There's another PR to update the parameter style -- should that be done here/now?  @cademack

howso/residuals.amlg
```diff
@@ -129,62 +191,173 @@
...
+			group_value_location (null)
+		)
+		(let
+			(assoc
+				group_values (list)
+				whole_shared_deviations_group
+					(call !GetSharedDeviationGrouping (assoc feature_group_to_retrieve feature_group_to_retrieve))
+			)
+			(if (> (size whole_shared_deviations_group) 0)
+				(map
```
howsohazard: Look at replacing the map/accum.

howso/feature_residuals.amlg
cademack: yep, those PRs are merged. So incoming changes to public labels and their parameters should have parameters documented accordingly (comments above each param)

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,202 @@
...
+	#!ValidateSharedDeviations
+	(let
+		(assoc error_messages (list))
+		;check to see if any features in a shared deviations group is a unique nominal
+		(if (size unique_nominals_set)
+			(declare (assoc
+				invalid_unique_shared_deviations_features
+					(filter
+						(lambda
+							(contains_value (indices unique_nominals_set) (current_index))
```
cademack: `(contains_value (indices ...))` can just be `(contains_index ...)`

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,202 @@
...
+						)
+						(get shared_deviations_map "shared_deviations_key_map")
```
cademack: nitpicky, but I think it would make sense to filter on the `(indices)` of `(get shared_deviations_map "shared_deviations_key_map")`. That way invalid_unique_shared_deviations_features would be a list like the var name sounds, then you wouldn't need to use (indices) in the error message code below

howso/attribute_maps.amlg
```diff
@@ -195,6 +195,202 @@
...
+		)
+	)
+
+	;Validates the necessary parameters for shared deviations, making sure that no shared deviations features are unique nominals.
+	#!ValidateSharedDeviations
+	(let
+		(assoc error_messages (list))
+		;check to see if any features in a shared deviations group is a unique nominal
+		(if (size unique_nominals_set)
+			(declare (assoc
```
cademack: wrapping up the pattern of: building a list of invalid features, checking its size, and conditionally accuming errors in a `(let (assoc ) ...)` may be easier to follow. 

Wouldn't have to keep the extra vars around either after the checks

howso/feature_residuals.amlg
```diff
@@ -310,6 +315,20 @@
 			(call !RunFullResiduals)
 		)
 
+		(if use_shared_deviations
+			(if (not (size !sharedDeviationsMap))
+				;if there are no set shared residuals, ignore shared deviations
+				(assign (assoc use_shared_deviations (false)))
+				(call !PrepSharedDeviations (assoc
+					features_list (assoc "features" features)
+					feature_residuals_lists (assoc " feature_residuals_lists" feature_residuals_lists)
```
cademack: looks like there's a space in the key here?

howso/feature_residuals.amlg
```diff
@@ -310,6 +315,20 @@
 			(call !RunFullResiduals)
 		)
 
+		(if use_shared_deviations
+			(if (not (size !sharedDeviationsMap))
+				;if there are no set shared residuals, ignore shared deviations
+				(assign (assoc use_shared_deviations (false)))
+				(call !PrepSharedDeviations (assoc
```
cademack: I'd like a new line between the positive/else conditional branches, makes it easier to read for me

cademack: otherwise my brain groups it together as 1 block

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,142 @@
...
 	)
 
+	;Helper function that takes a mapping or list that has been reduced, and expands it by readding the features from the groups that have been dropped.
+
+	;parameters:
+	; compressed_values: an assoc or list of values that need to expanded. If a list, adds back in the features from shared deviations groups that were removed
+	;  for calculations If an assoc, adds back in removed features as keys and copies the value of the primary key for each shared deviation group as its value.
+	;       e.g. {"a": 2.5, "b": 0} -> {"a": 2.5 "b":9 "c": 2.5 "d": 2.5} if features a, c, d were in a shared deviations group.
+	#!ExpandForSharedDeviations
+		(declare
```
cademack: unindent this method please

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,142 @@
...
+						;grab the values for the removed features by looking up their primary key values
+						(declare (assoc
+							removed_features_completed
+								(map
+									(lambda
+										(get compressed_values (get !sharedDeviationsMap (current_index)))
+									)
+									(zip removed_features)
+								)
+						))
```
cademack: can this declare be removed and that (map) and its innards just be moved into the 2nd arg of the (append) below?

of course the comment should come along

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,142 @@
...
+			(assoc current_shared_deviations_key (get !sharedDeviationsMap feature_group_to_retrieve))
+			(assign (assoc current_group
+				(filter
+					(lambda (= (current_value) current_shared_deviations_key) )
+					!sharedDeviationsMap
+				)
+
+			))
+
+			(indices current_group)
```
cademack: should probably just functionalize this as well so that no assign is needed

howso/residuals.amlg
```diff
@@ -129,62 +181,127 @@
...
+	#!ReduceSharedDeviationsMapping
+	(declare
+		(assoc map_to_reduce (null))
+		(let
+			(assoc
+				;If a feature is part of a shared deviations group, but is not the key, then drop
+				features_to_drop
+					(filter
+						(lambda (not (contains_value (values !sharedDeviationsMap (true)) (current_value) )))
+						(indices !sharedDeviationsMap (true))
```
cademack: what does `(indices ... (true))` do?

cademack: docs don't seem to list a second arg for (indices)

howso/residuals.amlg
```diff
@@ -129,62 +181,127 @@
...
+					)
+			)
+			(filter
+				(lambda (not (contains_value features_to_drop (current_index)) ))
+				map_to_reduce
+			)
```
cademack: can this just be a (remove)? 

(remove map_to_reduce features_to_drop)

howso/residuals.amlg
```diff
@@ -1237,6 +1355,13 @@
 				)
 		))
 
+		(if use_shared_deviations
+			(call !PrepSharedDeviations (assoc
+				features_list (assoc "features_with_nulls" features_with_nulls)
+				feature_residuals_keys_lists (assoc " feature_residuals_lists" feature_residuals_lists)
```
cademack: space here in this key too

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,142 @@
...
+			joined_original_features_residuals_list
+				(zip (first (values features_list)) (first (values feature_residuals_lists)))
+		))
+
+		;assoc of features and their residuals that are not in a shared deviations group
+		(declare
+			(assoc non_shared_deviations_feaures_list
+				(remove joined_original_features_residuals_list (indices !sharedDeviationsMap))
+			)
+		)
```
cademack: please adjust the styling of this declare

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,142 @@
...
+								feature_group (call !GetSharedDeviationGrouping (assoc feature_group_to_retrieve (current_index 2)))
+								current_feature (current_index 1)
+							)
+							(reduce
+								(lambda (append (previous_result) (current_value)) )
+								(map
+									(lambda (get joined_original_features_residuals_list (current_value)))
+									feature_group
+								)
+							)
```
cademack: we joked, but I do think this might be the exact same as a (apply "append" ...). I'd much prefer that imo (looks simpler to me at least) so please try it and change it to that if it works lol

howso/feature_residuals.amlg
```diff
@@ -1063,6 +1142,142 @@
...
+						)
+					)
+					;group primary features
+					(zip (values !sharedDeviationsMap (true)))
```
cademack: dont need to get unique values if they are gonna get zipped. Zipping will only keep an assoc with the unique keys anyway

howso/attribute_maps.amlg
howsoRes: simplify `(contains_value (indices unique_nominals_set) (current_index))` to `(contains_index unique_nominals_set (current_index))`

howso/generate_features.amlg
```diff
@@ -698,6 +700,15 @@
 		(map
 			(lambda (append
 				(assoc
+					"shared_deviations"
+						;only need the lags in one feature's attributes. Putting it in the first lag feature to reduce prep processing.
+						(if use_shared_deviations
```
howsoRes: since null is the default 'else' clause, you can probably simplify this down to 
(if (and use_shared_deviations (= 1 (current_value 1))) 
    (append ...)
)

howso/feature_residuals.amlg
```diff
@@ -1102,6 +1191,152 @@
...
+								(lambda
+									(not (contains_value
+										(indices compressed_values)
+										(current_value)
+									))
+								)
+								(indices !sharedDeviationsMap)
+							)
+					)
+					(assign (assoc
```
howsoRes: the pattern of `assign assoc compressed_values append compressed_values` an be simplified into an `accum assoc compressed_values`

howso/feature_residuals.amlg
```diff
@@ -1102,6 +1191,152 @@
...
+	; feature_residuals_lists: A features residuals whose lists are in the order of the features_list
+	#!PrepSharedDeviations
+	(declare
+		(assoc
+			features_list (assoc)
+			feature_residuals_lists (assoc)
+		)
+
+		(declare (assoc
+			joined_original_features_residuals_list (zip features_list feature_residuals_lists)
```
howsoRes: mis-named, this is a map not a list. it's a mapping of feature -> residual_list.  same with the non_shared one below

howso/feature_residuals.amlg
```diff
@@ -1102,6 +1191,152 @@
...
+	;Helper function to prep shared feature deviations, by taking in a list of features and their list of residuals,
+	; combining all of the residuals under the primary feature key of the group of the shared deviations, and removing the features
+	; from the features list whose residuals have been combined into another features.
+	;parameters:
+	; features_list: A features list
+	; feature_residuals_lists: A features residuals whose lists are in the order of the features_list
+	#!PrepSharedDeviations
+	(declare
+		(assoc
+			features_list (assoc)
```
howsoRes: rename this to `features` since that's what is passed in anyway, and is already a list of features, features_list sounds like it's a list of lists of features.

howso/feature_residuals.amlg
```diff
@@ -1102,6 +1191,152 @@
...
+				(map
+					(lambda
+						(let
+							(assoc
+								;the group of features in the shared deviations group
+								feature_group (call !GetSharedDeviationGrouping (assoc feature_group_to_retrieve (current_index 2)))
+								current_feature (current_index 1)
+							)
+							(apply "append"
+								(map
```
howsoRes: this entire map statement could be simplified into:
`(filter (unzip joined_original_features_residuals_list feature_group))`

howso/feature_residuals.amlg
```diff
@@ -1102,6 +1191,152 @@
...
+			features_list (assoc)
+			feature_residuals_lists (assoc)
+		)
+
+		(declare (assoc
+			joined_original_features_residuals_list (zip features_list feature_residuals_lists)
+		))
+
+		;assoc of features and their residuals that are not in a shared deviations group
+		(declare (assoc
```
howsoRes: this map isn't necessary at all, see comment re 'expanded_mapping' below

howso/feature_residuals.amlg
```diff
@@ -1102,6 +1191,152 @@
...
+			non_shared_deviations_features_list
+				(remove joined_original_features_residuals_list (indices !sharedDeviationsMap))
+		))
+
+		;assoc of primary shared deviations features and their residuals that are combined from all the other features in their shared deviations group
+		(declare (assoc
+			shared_deviations_residuals_list_combined
```
howsoRes: rename, should have _map in the variable name

howso/feature_residuals.amlg
```diff
@@ -1102,6 +1191,152 @@
...
+				;the combined list has a list for all feature deviations groups, this filters out groups with no values
+				(filter
+					(lambda (!= [] (current_value)))
+					shared_deviations_residuals_list_combined
+				)
+		))
+
+		;add the two assoc back together
+		(declare (assoc
```
howsoRes: you could create this mapping by taking `joined_original_features_residuals_list` (or whatever you rename it to), removing all indices from it that are !sharedDeviationsMap and then append `shared_deviations_residuals_list_combined` (or whatever you rename it to), that way you completely get rid of `non_shared_deviations_features_list`

howso/residuals.amlg
```diff
@@ -129,62 +181,127 @@
...
+											(call !GetSharedDeviationGrouping (assoc feature_group_to_retrieve (current_index 2)))
+											(null)
+										)
+									current_feature (current_index 1)
+								)
+								(declare (assoc
+									class_counts_map
+										(if (size feature_grouping)
+											;get the class counts for all the features in the shared deviations group
+											(reduce
```
howsoRes, line 196: this block of code does not work at all
a) this always results in a null, because if you use a reduce, the code needs to be in a (lambda) like other iterators
b) doing an (apply "append" will append the assocs, but if they have same indices/classes they will be clobbered, eg {a:5, b:4} appended with {a:3, c::2} results in {a:3, b:4, c:2} and not the expected {a:8, b:4, c:2}? nor {a:[5,3], b:4, c:2}?   we'll need to rethink this

howso/residuals.amlg
```diff
@@ -1308,12 +1446,56 @@
 										;compute from the existing feature max-min value or bounds if provided
 										(max
 											(-
-												(get !featureMarginalStatsMap (list weight_feature feature "max"))
-												(get !featureMarginalStatsMap (list weight_feature feature "min"))
+												(if (size feature_grouping)
+													(apply "max"
+														(filter
+															(lambda (!= (null) (current_value)))
```
howsoRes: (filter) by default only leaves non-null values, therefore there's no need for these (lambda) that match on nun-null, you can just do `(filter (map ... `

howso/residuals.amlg
```diff
@@ -129,62 +181,127 @@
...
+
+	;Takes a assoc of of values and filters the assoc to only features that either do not belong to a group of shared deviations, or are the primary
+	; key for a group of shared deviation features.
+	;
+	;parameters:
+	; map_to_reduce: An assoc that is to be reduced.
+	#!ReduceSharedDeviationsMapping
+	(declare
+		(assoc map_to_reduce (null))
+		(let
```
howsoRes: This whole method can be simplified into:
```
(remove map_to_reduce
    ;remove non-primary key features
    (remove 
        (indices !sharedDeviationsMap)
        ;primary keys
        (indices (zip (values !sharedDeviationsMap))
    )
)
```

Though since this pattern, as well as just having primary keys is going to be used over and over, it may make sense to cache them into model variables as well. perhaps !sharedDeviationsPrimaryFeatures and !sharedDeviationsNonPrimaryFeatures ? That we there's no need to do `(indices (zip (values !sharedDeviationsMap)))` every time you need access to just the primary features

howso/residuals.amlg
```diff
@@ -749,8 +841,44 @@
...
+						(declare (assoc
+							class_counts_above_stat_significance
+								(filter
+									(lambda (>= (current_value) nominal_class_statistical_significant_count))
+									(values (get !expectedValuesMap [weight_feature feature "class_counts"]))
+								)
+						))
+
+						(declare (assoc
+							smallest_count_above_statistical_significance (first (sort class_counts_above_stat_significance))
```
cademack, line 867: surely this can just be a (min)?

PR #275, opened 2024-09-03
21450: Fixes issues in time series where value features without nulls could synth derived features with nulls

howsoRes:
Changed the set of derived features (e.g., .value_delta_1) to be a map of the derived feature -> its parent feature, (e.g., { .value_delta_1 -> value }, for fast lookup, such that when derived features (e.g., '.value_delta_1') are derived, if their parent features (e.g., value) don't have nulls, it won't allow nulls for the derived feature either.

### DIFF
--- a/howso/react_series_utilities.amlg
+++ b/howso/react_series_utilities.amlg
@@ -1419,7 +1419,7 @@
 									(lambda (or
 										(= (false) (get !featureNullRatiosMap [(current_value 1) "has_nulls"]))
 										;rate or delta feature whose 'parent' feature has no nulls
-										(= (false) (get !featureNullRatiosMap [(get !derivedFeaturesSet (current_value 1)) "has_nulls"]))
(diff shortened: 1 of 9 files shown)


howso/react_series_utilities.amlg
```diff
@@ -62,6 +62,14 @@
         )
 
 		(if !inactiveFeaturesNeedCaching (call !UpdateInactiveFeatures))
+		;if for some reason expected values haven't been cached, do that here
+		(if (= 0 (size !expectedValuesMap))
+			(call !CacheExpectedValuesAndProbabilities (assoc
```
howsohazard, line 67: Make sure this method only does one assign to entities

cademack (approved):

cademack (dismissed):
test fail :(

howsohazard (approved):

PR #278, opened 2024-09-04
21470: Fixes missing action_values bug

jdbeel:


howso/react_series.amlg
```diff
@@ -317,6 +317,8 @@
 			context_values (null)
 			;{type "list" values "string"}
 			action_features (list)
+			;{type "list"}
```
cademack: this is the old "batch_react_series" so needs to be a 2d list. I'd just copy the one for context_values, unless you know something I don't

cademack (approved):

PR #279, opened 2024-09-04
21398: Updates the Amalgam unit test framework for ease of use

cademack:
Changes the unit test framework in the following ways:

- Adds strict checks for the `exp` variable for assert methods that use it. Calling these methods without setting the a non-null value for `exp` will now trigger a failure
- Updated the value printing logic to always print values from a failed assert if `#print_values_on_fail` is `(true)` which is on by default and switches the previous `#verbose_on_fail` which caused all assert values to be printed after a single failure to now be `(false)` by default.
- Parametrizes the sorting logic in `#assert_approximate`
- Some minor style/comment changes

Also updates the type schema and a unit test as "familiarity_conviction" is not actually a functional react detail.

### DIFF
--- a/unit_tests/unit_test.amlg
+++ b/unit_tests/unit_test.amlg
@@ -18,7 +18,10 @@
 	#num_failed_asserts 0
 
 	;if set to true will auto-output verbose on any fail
-	#verbose_on_fail (true)
+	#verbose_on_fail (false)
+
+	;if set to true, exp and obs values will be printed when assertions fail
+	#print_values_on_fail (true)
 
 	;name of unit test
 	#unit_test_name (null)
@@ -68,7 +71,7 @@
 	;helper function that gives failure output when a non-null value is not given for exp, skips check if called within a recursive assertion
 	#verify_exp
 	(if (and (= (null) exp) (not inner))
-		(call failed_assert_output (assoc msg "FAILED: A non-null value was not given for exp. If the expected value is (null), use #assert_null.\n"))
+		(call failed_assert_output (assoc msg "FAILED: A non-null value was not given for exp. If the expected value is (null), use #assert_null."))
 	)
 
 	; returns true (1) if the passed in numbers are within the specified percent of each other, otherwise returns false (0)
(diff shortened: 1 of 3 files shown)


unit_tests/unit_test.amlg
```diff
@@ -214,188 +238,234 @@
...
-			(print msg)
-			(if (= 2 verbose_mode)
-				(print " at\n" (stack)) ;TODO: figure out a less verbose/more concise way to give feedback as to where exactly the failure occured
-			)
+	(seq
+		(assign_to_entities (assoc num_failed_asserts (+ 1 num_failed_asserts)))
+		(assign (assoc failed_assert (true) ))
+		(print msg)
+		(if (= 2 verbose_mode)
+			(print " at\n" (stack)) ;TODO: figure out a less verbose/more concise way to give feedback as to where exactly the failure occured
```
howsohazard: TODOs need an associated ticket number

unit_tests/unit_test.amlg
```diff
@@ -37,33 +40,39 @@
...
+		(call assert_true (assoc obs (!= (null) unit_test_name)))
+		(call assert_false (assoc obs (= (null) unit_test_name)))
+		(call exit_if_failures (assoc quiet_pass (true) msg "Unit test framework Failed to load!"))
 
-			(print "=====================================\n")
-		)
+		(print "=====================================\n")
+	)
+
+	;helper function that gives failure output when a non-null value is not given for exp, skips check if called within a recursive assertion
```
howsoRes: 'gives' is a bit too generic, maybe '... that prints failure output ...' ?
also if this is a helper method that shouldn't be directly called by unit tests, make it private with a preceding !
also consider doing same for other helper/internal-only methods?

unit_tests/unit_test.amlg
```diff
@@ -214,188 +238,234 @@
...
 
 	;asserts that the obs parameter is null
 	; obs : item to check if null
 	#assert_null
-		(seq
-			(if (= (null) obs)
-				(print "PASSED (null)")
+	(declare
+		(assoc
+			obs "You did not pass in a value for obs" ;intentionally not null
```
howsoRes: comment on same line?

cademack: ah right

unit_tests/unit_test.amlg
cademack: just gonna remove the comment honestly, this doesn't seem necessary

howsoRes (approved):

PR #287, opened 2024-09-10
21502: Ignores early lag-dependent features for residuals computation for time series datasets

howsoRes:
TS flow creates a new internal-only feature called ".series_index" stores each cases's index with respect to its series.
During residual computation, lag features whose offset is larger than the series index of their case are not included, e.g, a lag feature of lag 5 will not be included when computing residuals for case with .case_index=3.

### DIFF
--- a/howso/residuals.amlg
+++ b/howso/residuals.amlg
@@ -323,7 +323,7 @@
 				(if !tsTimeFeature
 					(map
 						(lambda (or
-							;store value of 'max_row_la'g for lag features and 'ts_order' for delta or rate features
+							;store value of 'max_row_lag' for lag features and 'ts_order' for delta or rate features
 							(get (current_value) "max_row_lag")
 							(get (current_value) "ts_order")
 						))
--- a/howso/residuals.amlg
+++ b/howso/residuals.amlg
@@ -439,7 +439,7 @@
 		;keep a copy of all originally specified features
 		(assign (assoc
 			case_features
-				(values (append context_features features) (true))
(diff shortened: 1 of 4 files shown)


howso/residuals.amlg
```diff
@@ -473,8 +473,21 @@
 						)
 					)
 				)
+			lag_feature_amount_map (null)
```
cademack: move the assign value logic before here

howso/residuals.amlg
```diff
@@ -317,8 +317,29 @@
 			p_parameter (get hyperparam_map "p")
 			dt_parameter (get hyperparam_map "dt")
 			query_feature_attributes_map (get hyperparam_map "featureDomainAttributes")
+
+			;store an assoc of lag/rate/delta feature -> lag/order amount for time series flows
+			ts_feature_lag_amount_map
+				(if !tsTimeFeature
+					(map
+						(lambda (or
+							;store value of 'max_row_la'g for lag features and 'ts_order' for delta or rate features
```
cademack: typo here

howso/residuals.amlg
```diff
@@ -317,8 +317,29 @@
...
+						(filter
+							(lambda (get (current_value) "ts_type") )
+							!featureAttributes
+						)
+					)
+				)
 		))
 
+		(if !tsTimeFeature
+			(assign (assoc max_lag_index_value (apply "max" (values ts_feature_lag_amount_map)) ))
```
cademack, line 340: should this be a declare rather than an assign?

cademack (dismissed):

howsohazard (approved):

PR #291, opened 2024-09-11
21442: Adds label annotations that are parsed and used in get_api output

cademack:
Adds logic to `#get_api` that adds "idempotent" and "long-running" flags to the object representing each method available on the Howso entity.

The idempotent flag defaults to True and the long-running flag default to False. To override these default values, add a commented (assoc) within the comments above the label specifying the overriding values.

To reviewers, please verify the accuracy of the flags intentionally specified and suggest any other public labels that may need to be labeled as *not* idempotent or slow-running


howso/conviction.amlg
```diff
@@ -2,6 +2,7 @@
 (null
 
 	;computes various data, such as familiarity convictions and distance contribution for each case in the model and stores them into specified features.
+	;{long-running (true)}
```
howsohazard: also not idempotent

howso.amlg
```diff
@@ -769,34 +774,68 @@
...
+									;idempotent is true by default
+									"idempotent"
+										(if (contains_index label_flags_map "idempotent")
+											(get label_flags_map "idempotent")
+
+											(true)
+										)
+									;long_running is false by default
+									"long_running"
+										(if (contains_index label_flags_map "long_running")
```
howsoRes, line 792: for consistency, this is fine. though you could just do `(or (get label_flags_map "long_running"))` and it'll work the same since (or (null)) evals to (false)

howso/analysis.amlg
```diff
@@ -1,6 +1,7 @@
 ;Contains methods for hyperparameter analysis and feature deviation and weights calculations.
 (null
 	;automatically analyze the model using stored parameters from previous analyze calls
+	;{long_running (true) idempotent (false)}
```
howsoRes, line 4: read_only (false) as well

howso/react_aggregate.amlg
```diff
@@ -2,6 +2,7 @@
 (null
     ;Computes, caches, and returns specified details and feature prediction statistics such as Mean Decrease in Accuracy (MDA), residuals (accuracy, Mean Absolute Error),
     ; precision, recall, etc. Returns details and feature prediction stats for all features in the format of feature -> assoc stat -> value
+	;{long_running (true)}
```
howsoRes, line 5: probably idempotent is false since i think you'll get diff results everytime, esp if asking for robust residuals and such?

howso.amlg
cademack: I see, gonna keep it this way for the pattern/readability though

howso/analysis.amlg
cademack: switched default to false so it should be ok

howsoRes (dismissed):

howsoRes (approved):

PR #292, opened 2024-09-12
21517: Updates `#edit_cases` to re-derive non-time-series features that auto-derive on train, also fixes some bugs related to derivation

cademack:
Adds logic to `#edit_cases` that will check if any non-time-series derived features within the Trainee are derived from features that were edited. If there are any, they are re-derived with the updated values and the cases are updated with the new values.

### DIFF
--- a/howso/update_cases.amlg
+++ b/howso/update_cases.amlg
@@ -218,17 +218,103 @@
 		))
 
 		(if case_ids
-			(map
-				(lambda
-					(call !UpdateCaseWithHistory (assoc
-						case_id (current_value 1)
-						features features
-						feature_values feature_values
-						unencoded_feature_values unencoded_feature_values
-						session session
-					))
+			(seq
+				(map
+					(lambda
+						(call !UpdateCaseWithHistory (assoc
+							case_id (current_value 1)
+							features features
+							feature_values feature_values
+							unencoded_feature_values unencoded_feature_values
+							session session
+						))
+					)
+					case_ids
 				)
-				case_ids
+
+				;update derived features if necessary
+				(let
+					(assoc
+						affected_derived_features
+							;any features edited that are the source feature of a non-TS derived feature
+							(filter
+								(lambda
+									(and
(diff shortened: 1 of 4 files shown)


howso/update_cases.amlg
```diff
@@ -218,17 +218,102 @@
...
+									overwrite (true)
+								))
+							)
+						)
+						affected_derived_features
+					)
+
+					;update features if necessary so that data mass change is correct
+					(if (size affected_derived_features)
+						(assign (assoc features (append features affected_derived_features) ))
```
howsoRes: any reason why this isn't an accum instead of an assign-append ?

howso/update_cases.amlg
```diff
@@ -218,17 +218,102 @@
...
+									label_name derived_feature
+									overwrite (true)
+								))
+							)
+						)
+						affected_derived_features
+					)
+
+					;update features if necessary so that data mass change is correct
+					(if (size affected_derived_features)
```
howsoRes: no need to do the size check here, just do the accum below since `affected_derived_features` should be an empty list and will accum correctly ?

howso/update_cases.amlg
```diff
@@ -218,17 +218,102 @@
...
+										(=
+											[0]
+											(values
+												(get_all_labels (parse (get !featureAttributes (list (current_value 1) "auto_derive_on_train" "code"))))
+												(true)
+											)
+										)
+									)
+								)
+								(indices !derivedFeaturesMap)
```
howsoRes: you could (or) this with an [ ] to ensure that affected_derived_features is always a list and never a null.

cademack: sure

howso/update_cases.amlg
cademack: I was concerned about it being added as a sublist or something like that... but its fine, will do

howsoRes (approved):

PR #308, opened 2024-09-25
21028: Improves performance of creation of SDM by only using classes in the sparse confusion matrix and not all classes in model during computation

howsoRes:


howso/hyperparameters.amlg
```diff
@@ -1100,7 +1089,24 @@
 			(conclude (null))
 		)
 
-		(declare (assoc num_classes (size all_classes_map) ))
+		(declare (assoc num_classes (size (get !expectedValuesMap [weight_feature feature "class_counts"])) ))
```
cademack, line 1092: This check and compute if empty pattern is here twice, should we instead just assign it to the cache after the first compute so it will definitely be present later?

cademack (dismissed):

howso/hyperparameters.amlg
howsoRes: no, that's what the code used to do by caching all_cases_map up here, but it wasn't efficient since it could be a very large assoc that uses up memory, and we don't want to be writing to the entity here.

a) the second use should be very rare, this first/primary use is to get the number of classes and won't store the full assoc in memory. 
b) In flows where analyze happens it will just pull from !expectedValuesMap in both uses anyway
c) the primary/first use only needs the number of cases, which will later be optimized in Amalgam via a single query, the second use actually needs the full map, so later these queries will actually be different.

cademack (approved):

PR #309, opened 2024-09-26
21661: Adds descriptions for properties of FeatureAttributes, ReactDetails, and other schemas

cademack:
Adds an enormous amount of property descriptions across all of the typing schemas

### DIFF
--- a/howso/typing.amlg
+++ b/howso/typing.amlg
@@ -547,7 +547,7 @@
                     (assoc
                         influential_cases
 							(assoc
-								type  "boolean"
+								type "boolean"
 								description
                                     (concat
                                         "When true, outputs the most influential cases and their influence weights based on the surprisal of each case "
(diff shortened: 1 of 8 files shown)


howso/typing.amlg
```diff
@@ -147,53 +539,427 @@
...
+                                    (concat
+                                        "When true, outputs similarity conviction for the reacted case. Uses both context and action feature values as "
+                                        "the case values for all computations. This is defined as expected (local) distance contribution divided by "
+                                        "reacted case distance contribution."
+                                    )
+							)
+                        outlying_feature_values
+							(assoc
+								type  "boolean"
+								description
```
fulpm, line 665: Missing a description here

howso/typing.amlg
```diff
@@ -147,53 +539,427 @@
...
+                                        "computations. 'selected_prediction_stats' controls the returned prediction stats."
+                                    )
+							)
+                        selected_prediction_stats
+                            (assoc
+                                type "list"
+                                values
+                                    (assoc
+                                        type "string"
+                                        enum (get_value !supportedPredictionStats)
```
fulpm: Could this enum be a schema instead? Might be easier to use from a typing perspective.

fulpm: Also looks like it might be used in react aggregate details? Should probably reference it there too, currently that is just a freeform string.

howso/typing.amlg
cademack: oh good eye

howso/typing.amlg
```diff
@@ -1001,5 +1130,19 @@
                         "to be a new case (that is, a case not found in original dataset.)"
                     )
             )
+        SelectedPredictionStats
```
fulpm, line 1133: Lets just call it `PredictionStats`

fulpm (dismissed):

howsoRes (dismissed):

howsoRes (approved):

PR #311, opened 2024-09-30
21681: Adds support for required parameters and updates parameter types to include "null" where appropriate

cademack:
Also adjusts some outdated tests to not give invalid parameters.


howso/typing.amlg
```diff
@@ -1149,6 +1149,27 @@
                         "If all, then returns all including the confusion_matrix."
                     )
             )
+        DesiredConviction
+            (assoc
+                type ["number" "null"]
+                min 0.000001
```
fulpm: Might be better to have an `exclusiveMin` and `exclusiveMax` to handle this better. Then you could have `exclusiveMin: 0` for values > 0

fulpm (approved):

PR #316, opened 2024-10-02
21723: Updates initial default parameters 

howsoRes:
Also changes initial nominal deviation logic and reduces memory load during residual computation for nominal features

### DIFF
--- a/howso.amlg
+++ b/howso.amlg
@@ -630,7 +630,7 @@
 			!defaultHyperparameters
 				(assoc
 					"k" 8
-					"p" 0.1
+					"p" 1
 					"dt" -1
 					"featureWeights" (null)
 					"featureDeviations" (null)
--- a/howso.amlg
+++ b/howso.amlg
@@ -269,10 +269,10 @@
 	#initialize
 	(declare
 		(assoc
-			;{type "string"}
+			;{type "string" required (true)}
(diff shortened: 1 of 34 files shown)


cademack (dismissed):

howso/analysis_weights.amlg
```diff
@@ -235,9 +235,14 @@
 				initial_residual_values_map
 					(map
 						(lambda
-							;gap is always 1 for nominals, set initial value to gap / 2
+							;for nominals, random chance of getting a prediction wrong is: 1 - 1/num_classes
+							;set initial value to (1 - 1/num_classes) / 2
 							(if (contains_index !nominalsMap (current_index))
-								0.5
+								;0.5 - 0.5/num_classes
```
howsohazard: What is this comment doing?  Not exactly clear from above, and not exactly clear with code below.

howso.amlg
```diff
@@ -630,7 +630,7 @@
 			!defaultHyperparameters
 				(assoc
 					"k" 8
-					"p" 0.1
+					"p" 1
```
howsohazard: Does this mean there are always deviations?  What happens if you use it without running analyze?  We don't want just p=1 in that case.

howsohazard (approved):

PR #322, opened 2024-10-10
21355: Defines return-type schemas for public labels

cademack:
Defines return-type schemas for public labels. These return types are in comments above the (assoc) node of public labels.

#get_api parses these comments and adds them to the output objects for each label. Additionally, unit_test_howso.amlg has been updated to inject code into #!Return that checks the return payloads against the type schema and errors out on failure.


howso/hierarchy.amlg
```diff
@@ -5,14 +5,26 @@
...
 	;Returns the full entity path to a child trainee provided its unique trainee id if it is contained in the hierarchy.
 	;Iterates down the hierarchy searching for a trainee that matches the specified id, returns null if not found or
 	;a string error if found but trainee is stored externally as an independent trainee.
 	;{read_only (true) idempotent (true)}
 	#get_entity_path_by_id
 	(declare
+		;returns {
+		; 	type ["list" "string" "null"]
+		; 	values "string"
+		; 	description "The full entity path to the child trainee, a string describing an error, or null if the child trainee was not found"
```
fulpm, line 26: The return string could be an error message? Instead of using the standard errors key outside of payload?

cademack: unfortunately accurate

fulpm (dismissed):

howsoRes (approved):

PR #323, opened 2024-10-11
21806: Fixes react_series output detail formats

howsoRes:


howso/details.amlg
```diff
@@ -830,6 +830,10 @@
 				(list)
 			)
 		)
-		(zip details)
+		(remove
```
howsohazard, line 836: Why is this here?  Seems that either the method needs a comment or something.

howso/distances.amlg
```diff
@@ -149,7 +149,7 @@
 		(accum (assoc
 			output
 				(assoc
-					distance_ratio
+					"distance_ratio"
```
howsohazard, line 152: why the quotes?

howsohazard: same for the changes below

howso/react_series_utilities.amlg
```diff
@@ -179,7 +179,11 @@
 									(map (lambda (get (current_value) detail)) series_output)
 								)
 							)
-							(zip requested_details)
+							(remove
```
howsohazard, line 183: This again seems weird.

howso/distances.amlg
howsoRes: the quotes are because this is an output dictionary for the user, not declaring variables on the stack.  That's my preferred method of denoting the difference, an assoc with non-quoted keys are variables, quoted keys are simply keys for some data. Also it makes it consistent with the other detail outputs, this code was not consistent before. same for the others.

howsohazard (approved):

PR #325, opened 2024-10-11
21637: Adds thresholds to auto-ablation, MAJOR

jdbeel:
Also updates the parameters for auto-ablation and reduce-data to remove mention of model size.

### DIFF
--- a/howso/hyperparameters.amlg
+++ b/howso/hyperparameters.amlg
@@ -936,6 +936,9 @@
 				auto_ablation_weight_feature !autoAblationWeightFeature
 				minimum_model_size !autoAblationMinModelSize
 				influence_weight_entropy_threshold !autoAblationInfluenceWeightEntropyThreshold
+				abs_threshold_map !autoAblationAbsThresholdMap
+				delta_threshold_map !autoAblationDeltaThresholdMap
+				rel_threshold_map !autoAblationRelThresholdMap
 				exact_prediction_features !autoAblationExactPredictionFeatures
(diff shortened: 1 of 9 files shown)


howso/hyperparameters.amlg
```diff
@@ -992,6 +995,24 @@
 			;{type "number"}
 			;the number of ablated cases to compute influence weights for distribution at a time
 			ablated_cases_distribution_batch_size 100
+			;{ref "AblationThresholdMap"}
+			;a map of measure names (any of the prediction stats) to a map of feature names to threshold value.
+			; absolute thresholds will cause data reduction to stop when any of the measure values for any of
+			; the features for which a threshold is defined go avove the threshold (in the case of rmse and
```
howsoRes: several avove instead of above instances

howso/train.amlg
```diff
@@ -705,11 +705,52 @@
...
+					)
+				prev_prediction_stats_map {}
+				new_prediction_stats_map {}
+				thresholds_satisfied (false)
 			))
 
 			;split by batches of cases until next analyze
 			(while (< input_case_index (size cases))
 
+				(if (and thresholds_enabled !autoAblationEnabled)
```
howsoRes: !autoAblationEnabled or (not skip_ablation) ? ie, I don't think should this block be run during the first training batch before any cases have even been trained, or when only a few cases have been trained.

cademack (dismissed):

howsoRes (dismissed):

cademack (dismissed):

howsoRes (dismissed):

howsohazard (approved):

PR #329, opened 2024-10-15
21830: Moves larger return-type schemas into a separate module and adjusts `#get_api` accordingly

cademack:


jdbeel:
Should we add a typing directory which contains the typing modules?

fulpm (dismissed):

fulpm (dismissed):

howsoRes (approved):

PR #330, opened 2024-10-15
21710: Refactors hyperparameter map and removes .targetless, MAJOR

howsoRes:
- refactors hyperparameter map to contain 'targetless' and/or 'targeted', which contain their respective maps
- no more robust/full layer in hyperparameter map
- no more .targetless


howso/hyperparameters.amlg
```diff
@@ -596,23 +566,20 @@
...
 		; 		"default_hyperparameter_map" {ref "HyperparameterMap" description "The map of default hyperparameters"}
 		; 		"auto_analyze_enabled" {type "boolean" description "Flag indicating of auto-analysis is enabled."}
 		; 		"analyze_threshold" {type "number" description "The number of cases at which the auto-analysis is triggered."}
 		; 		"analyze_growth_factor" {type "number" description "The scalar rate at which the number of cases to trigger an auto-analysis grows with each iteration"}
 		; 	}
 		; }
 		(assoc
-			;{type "string"}
-			;the target feature of the desired hyperparameters
+			;{type ["string" "boolean"]}
```
howsohazard: Can't have a nullable string?

howso/react_aggregate.amlg
```diff
@@ -171,11 +171,11 @@
 			;target feature for which to do computations. If "prediction_stats_action_feature" and "feature_influences_action_feature"
 			;	are not provided, they will default to this value.
 			action_feature (null)
-			;{type "string"}
+			;{type ["string" "boolean"]}
```
howsohazard: also nullable string here?

howso/react_aggregate.amlg
```diff
@@ -171,11 +171,11 @@
...
 			;When calculating residuals and prediction stats, uses this target features's hyperparameters. The trainee must
 			;	have been analyzed with this feature as the action feature first. If both "prediction_stats_action_feature" and "action_feature" are not provided,
-			;	by default residuals and prediction stats uses ".targetless" hyperparameters. If "action_feature" is provided, and this value is not provided, will
-			;	default to the value of "action_feature".
+			;	or if this value is set to false, by default residuals and prediction stats uses "targetless" hyperparameters. If "action_feature" is provided, but
```
howsohazard: Improve wording

howso/hyperparameters.amlg
```diff
@@ -596,23 +566,20 @@
...
+			;the target feature of the desired hyperparameters, if false will output targetless hyperparameters
```
howsohazard: should describe what happens if it is null, which it is by default

howso/hyperparameters.amlg
cademack: if the parameter isn't specified as "required", then null will be accepted. Though (null) has some meaning in the context of this parameter, so I agree we should probably add it to the list of accepted types here

howso/feature_residuals.amlg
```diff
@@ -55,19 +54,9 @@
 			param_path
```
cademack, line 54: This `param_path` variable/declaration may have been for when these residuals were stored in their own map. Check and remove if this is unused?

howso/hyperparameters.amlg
```diff
@@ -512,70 +517,35 @@
 				context_key_options
 					(sort
 						(lambda (> (size (current_value)) (size (current_value 1))))
-						(indices (get !hyperparameterMetadataMap feature) )
+						(indices (get !hyperparameterMetadataMap target_mode) )
```
cademack, line 527: If this returns the first candidate context key it finds to be suitable, then perhaps we should instead sort the keys in *decreasing* length. Maybe it already does that and the comment is wrong, I don't remember exactly how the lambda in sort behaves

howsohazard (dismissed):

howso/hyperparameters.amlg
```diff
@@ -689,86 +657,46 @@
...
 			(let
-				(assoc param_paths (list))
+				(assoc
+					param_paths []
+					keys []
+				)
+
+				;populate all provided param_paths from the specified hyperparameter_map by appending all the indices recursively
+				#!RecursivelyAccumulateParamPaths
 				(map
```
cademack, line 686: I think we should throw an (and) on this condition and also include dt and p. Mostly because I don't think it's too crazy to imagine a dataset with a feature named "k"

cademack: Not to say there isn't one that could also have "p" and "dt", but certainly less likely.

howso/feature_residuals.amlg
howsoRes: still being used

cademack (dismissed):

jackx111 (approved):

PR #339, opened 2024-10-22
21928: Updates applicable labels to automatically use populated case weights

cademack:
react_aggregate and analyze could be triggered in ways that left case-weight usage unspecified, but would not automatically use populated case weights from ablation/reduction. This fixes that

### DIFF
--- a/howso/react_aggregate.amlg
+++ b/howso/react_aggregate.amlg
@@ -223,7 +223,7 @@
 			(assign (assoc sample_model_fraction (null)))
 		)
 
-		;if not using case weights, change weight_feature to '.none'
+		;update case weight related parameters based on use_case_weights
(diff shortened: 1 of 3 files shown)


howso/react_aggregate.amlg
```diff
@@ -223,11 +223,47 @@
...
-		;if not using case weights, change weight_feature to '.none'
-		(if (not use_case_weights)
-			(assign (assoc weight_feature ".none"))
+		;update case weight related parameters based on use_case_weights
+		(if (= (false) use_case_weights)
+			(assign (assoc
+				weight_feature ".none"
+			))
+
+			(= (true) use_case_weights)
```
howsoRes: just bare `use_case_weights` should work fine.

howsoRes (approved):

PR #342, opened 2024-10-29
19790: Fixes time series ablation to properly derive values prior to ablation

howsoRes:


howso/train_ts_ablation.amlg
```diff
@@ -0,0 +1,562 @@
...
+			one_id (= 1 (size id_indices))
+			id_index (first id_indices)
+			needs_grouping (false)
+		))
+
+		(if one_id
+			(let
+				(assoc
+					unique_ids_set
+						(if one_id
```
cademack: one_id is already a truthy here

howso/train_ts_ablation.amlg
```diff
@@ -0,0 +1,562 @@
...
+				)
+			)
+		)
+
+		;output true if data needs grouping
+		needs_grouping
+	)
+
+
+	;hepler method
```
cademack: fun typo here

howso/train_ts_ablation.amlg
```diff
@@ -0,0 +1,562 @@
...
+	;uses cases and id_indices to output cases grouped by ids in columns specified by id_indices
+	#!GroupDataByIds
```
cademack, line 130: is this really better than just a multi column sort?

howso/train_ts_ablation.amlg
```diff
@@ -0,0 +1,562 @@
...
+				)
+			)
+		)
+
+		;return trained case_ids
+		output_case_ids
+	)
+
+	;derive then train with ablation
+	#!TrainTimeSeriesWithAblation
```
cademack: This method name is like the exact same as the other... If I am understaning the code correctly, I suggest we rename this one to something like #!TrainSingleSeriesWithAblation

howso/train_ts_ablation.amlg
```diff
@@ -0,0 +1,562 @@
...
+			series_ordered_by_features (get !featureAttributes [(first lag_features) "auto_derive_on_train" "ordered_by_features"])
+			needs_grouping
+				(call !DoesDataNeedGroupingById (assoc
+					data cases
+					id_indices id_indices
+				))
+		))
+
+		(if needs_grouping
+			(assign (assoc cases (call !GroupDataByIds) ))
```
cademack, line 196: Why must we put all the cases in a single list after getting a sublist for each series earlier? Then breaking it back into calls with lists of cases for each series?

Is it so we can track the training indices properly still?

howso/train_ts_ablation.amlg
howsoRes: it is only truthy by default, if user specifies multiple id_indices when calling this method, it'll be false.

howso/train_ts_ablation.amlg
howsoRes: Well, it's grouping without sorting, because it's not comparing the values to see which should go first or second, simply arranging them so they are lumped together in whatever order they happen to be. not sure it's better or worse, we can discuss.

howso/train_ts_ablation.amlg
howsoRes: ideally we won't break up that list, it's only done if grouping was necessary

howso/train_ts_ablation.amlg
howsoRes: if we sort, it will actually sort the groups by ids, whereas current code leaves the groups in the same order as they are, so if there are minimal changes that need to happen, my code will disturb it the least. not sure if it's slower than sort.

howso/train_ts_ablation.amlg
```diff
@@ -0,0 +1,596 @@
+(null
```
howsohazard, line 1: Methods need better documentation, and has TODO

howsohazard (approved):

howsohazard (approved):

PR #344, opened 2024-10-31
22017: Add readonly option to analyze

dmaze:


howso/analysis.amlg
```diff
@@ -14,6 +14,45 @@
 		)
 	)
 
+	;Commit the result of a previous analyze call
+	#commit_analyze
+	(seq
```
howsoRes, line 19: should be a declare-assoc block like all others.

howso/analysis.amlg
```diff
@@ -14,6 +14,45 @@
...
+				!defaultHyperparameters default_hyperparameters
+			))
+		)
+		(if (!= (null) hyperparameter_metadata_map)
+			(assign_to_entities (assoc
+				!hyperparameterMetadataMap hyperparameter_metadata_map
+			))
+		)
+		(accum_to_entities (assoc !revision 1))
+	)
```
howsoRes, line 55: last statement in a method must be a (call !Return)

howso/analysis.amlg
```diff
@@ -314,9 +360,23 @@
 			(call !RestoreHeldOutCases (assoc holdout_entity_name holdout_entity_name))
 		)
 
-		(accum_to_entities (assoc !revision 1))
-
-		(call !Return (assoc warnings (if (size analyze_warnings) (indices analyze_warnings)) ))
+		(let
```
howsoRes, line 363: change the flow of this so the code is:
```
(if readonly
    (call !Return (assoc warnings ... )
    ;else 
    (let 
       ...
       (call commit_analyze ...)  
       (call !Return ... )
    )

```

PR #348, opened 2024-11-05
22117: Enables Dynamic Deviations for generative reacts and details, and fixes an issue when using ordinal features with Dynamic Deviations

cademack:
Adds the necessary logic to utilize Dynamic Deviations for generative reacts and other react details where it previously would not be used. 

Additionally fixes an issue in Dynamic Deviations where ordinal feature values would be counted as continuous, which would consequently break the trained values and analyzed hyperparameters of the subtrainee.

Lastly, I also fix the performance tests to specify the filepaths correctly. Without the correct filepath, the Trainee could not properly create subtrainees.

### DIFF
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -315,7 +315,7 @@
 
 		(if (!= (null) result)
 			(seq
-				(call_entity trainee_path "initialize" (assoc trainee_id trainee_id)))
+				(call_entity trainee_path "initialize" (assoc trainee_id trainee_id))
 
 				(call !AddChildTraineeReferences (assoc
 					child_id trainee_id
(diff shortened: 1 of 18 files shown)


howso/hierarchy.amlg
```diff
@@ -315,7 +315,7 @@
 
 		(if (!= (null) result)
 			(seq
-				(call_entity trainee_path "initialize" (assoc trainee_id trainee_id))
+				(call_entity trainee_path "initialize" (assoc trainee_id trainee_id)))
```
howsoRes: this is an extra ) parethsesis which may be causing some issues

howsoRes (approved):

PR #354, opened 2024-11-11
13693: Adds "boundary_values" as a detail to #react, MINOR

cademack:
This PR adds a new detail to `#react`: "boundary_values_context_features". This detail allows users to ask questions about how much context values would have to change to observe differences in the resulting predictions.

The API changes are as follows:

- `#react` (and `#single_react`) now accept 2 additional indices within the `details` (assoc) parameter. 
- The first new index is "boundary_values_context_features" which takes a list of feature names to compute boundary values for. 
- The second new index is `boundary_values_action_outcome`, which accepts a map of conditions for each action feature. This allows users to specify a condition that indicates the boundary in which they are interested in. For example, a user could specify `{"feat1" [5 15]}` which would then find the boundary values that would cause the prediction of "feat1" to be within the range of 5 to 15. (This condition formatting matches that of the standard `Condition` schema)
- The ReactResponse object has a new key, "boundary_values", which holds a map of feature names to their computed boundary values.
   - For continuous and ordinal features, a tuple of boundary values is given allowing for a boundary to be found on either side of the given context value.
   - For nominal features, a single value is given. This value is the "most similar" class to the given context value that crosses the boundary or maximizes the change in the action.

When `boundary_values_action_outcome` is specified, the boundary is considered "well defined", and the algorithm will attempt to find the closest values to the given context values that cross the boundary. When it is unspecified, there are two possibilities. If all action features are nominal, then the boundary is still considered "well defined" and any change in the action values will indicate a crossed boundary and the algorithm still searches for the closest values that cross this boundary. In the case that `boundary_values_action_outcome` is unspecified and there are non-nominal action features, then the algorithm will attempt to find the closest local maxima of the gradient in the change of action values with respect to the change in context values.

With the current implementation, boundary values can be computed for continuous, nominal, and ordinal (numeric and string) features. It is not supported for edit-distance features (JSON, YAML, Amalgam).


unit_tests/ut_h_boundary_values.amlg
```diff
@@ -0,0 +1,120 @@
...
+	))
+
+	(print "Greater Bound: ")
+	(call assert_null (assoc
+		obs (get result ["boundary_values" 0 "ord" 1])
+	))
+
+
+
+	(- 22.1)
```
howsoRes: this test needs some more development / detail

howso/boundary_values.amlg
```diff
@@ -0,0 +1,814 @@
...
+								)
+								boundary_action_condition
+							)
+						))
+					)
+
+					;no specified boundary, define a function that will indicate a boundary based on action values
+					(let
+						(assoc
+							num_nominal_features (size (filter (lambda (= "nominal" (get !featureAttributes [(current_value 1) "type"]))) action_features))
```
howsoRes: (contains !NominalsMap (current_value)) is simpler and faster than checking 2 layers deep into attributes and comparing to a string.

howso/boundary_values.amlg
```diff
@@ -0,0 +1,814 @@
...
+					;now search the increasing size neighborhoods for similar values
+					(let
+						(assoc
+							case_bandwidth (get hyperparam_map "k")
+							similar_cases_lists (null)
+						)
+
+						(while
+							(and
+								(< (current_index) 3)
```
howsoRes, line 401: please comment what this 3 layer deep "context" for current_index is

howso/boundary_values.amlg
```diff
@@ -0,0 +1,814 @@
...
+								(!=
+									;the unique classes of the feature observed in the neighborhood
+									(size (values (get similar_cases_lists 2) (true)))
```
howsoRes, line 404: (last similar_cases_lists) instead of (get .. 2)

howso/boundary_values.amlg
```diff
@@ -0,0 +1,814 @@
...
+							(apply "round" (append (/ (+ left right) 2) rounding_tuple))
+							(/ (+ left right) 2)
+						)
+				))
+
+				;binary search until within a residual
+				(while
+					(>
+						(- right center)
+						; (get hyperparam_map ["featureResiduals" query_feature])
```
howsoRes: should this be a TODO?

cademack: err just missed flipping this back while testing

unit_tests/ut_h_boundary_values.amlg
cademack: yeah this commit of it was very work-in-progress

howso/boundary_values.amlg
cademack: will do

howso/boundary_values.amlg
cademack: will do

howso/boundary_values.amlg
```diff
@@ -0,0 +1,879 @@
...
+											)
+											;right difference greater, return index to the left
+											(- index_of_max_boundary_ratio 1)
+
+											;else right
+											(+ index_of_max_boundary_ratio 1)
+										)
+									)
+							))
+
```
howsoRes, line 530: add a comment explaining what this pair is

howso/boundary_values.amlg
```diff
@@ -0,0 +1,879 @@
...
+								(contains_value !ordinalFeatures query_feature)
+								;Ordinals Block
+								(call !ComputeOrdinalBoundaryValues)
+
+
+								(contains_index !editDistanceFeatureTypesMap query_feature)
+								;Not supported.
+								(null)
+
+								;else will assume continuous... TODO: this probably isn't exactly correct, the not of all previous conditions might not imply its continuous
```
howsoRes: either need ticket or remove todo

cademack: I think this is probably correct now. Put this comment in before other blocks were added

howsoRes (approved):

PR #356, opened 2024-11-15
20486: Shared Deviations with Time-series tweaks, MINOR

cademack:
This PR adds the implementation for shared deviations (not done by me, but inherited from others branches), and adds logic to increase sampling when using TS and Shared Deviations that ensure the desired/expected number of non-null samples for a feature is achieved by scaling the number of samples taken in reference of how many of the shared deviation features are null for the average case.

For example, if there is a shared deviations feature group with 40 features and Analysis would use 2,000 samples, then we would normally expect to get 80,000 samples for the group of features in total. However, in the case of time-series, shared deviations group will contain lags and many of these will be null for many cases, this PR adds logic that estimate the mean number of non-null values for each shared deviation group and finds the estimated number of samples needed to get the desired number of non-null samples for the shared deviation group that needs the most.


howso/analysis_weights.amlg
```diff
@@ -204,6 +212,74 @@
...
+		(declare (assoc
+			avg_non_null_values_per_case_per_group
+				(map
+					(lambda
+						(floor (/
+							(apply "+"
+								(map
+									(lambda
+										;the number of non-null values in the case of features in the shared deviation group
+										(size (filter (retrieve_from_entity (current_value) (current_value 1))))
```
howsoRes, line 252: (retrieve_from_entity) only works on a single case, so (size (filter (retrieve_from_entity... )) will always return either 1 or 0
I think you meant to put the (size (filter  outside of the map that's iterating over sample_cases ?

howso/analysis_weights.amlg
```diff
@@ -204,6 +212,74 @@
...
+		;take a sample of 1000 cases and get the average number of non_null values for features in each shared deviation group
+		(declare (assoc
+			avg_non_null_values_per_case_per_group
+				(map
+					(lambda
+						(floor (/
+							(apply "+"
+								(map
+									(lambda
+										;the number of non-null values in the case of features in the shared deviation group
```
howsoRes, line 251: add a comment that (current_value 1) is the list of all the features in the shared deviation group

howsoRes (approved):

PR #359, opened 2024-11-20
22254: Renames set_contained_trainee_maps to set_hierarchy_relationships, MAJOR

howsoRes:
also adds get_hierarchy_relationships() and get_parent_id()

### DIFF
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -873,7 +873,7 @@
 		;	additional_indices (false)
 		; 	indices {
 		; 		"parent_id" {
-		; 			type "string"
+		; 			type ["string" "null"]
 		; 			description "The unique string identifier for the parent of the trainee."
 		; 		}
 		; 		"path_to_id_map" {
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -874,7 +874,7 @@
 		; 	indices {
 		; 		"parent_id" {
 		; 			type "string"
-		; 			description "The unique string identifier for the parent of the tarinee."
+		; 			description "The unique string identifier for the parent of the trainee."
 		; 		}
 		; 		"path_to_id_map" {
 		; 			type "assoc"
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -816,7 +816,7 @@
 	#set_hierarchy_relationships
 	(declare
 		(assoc
-			;{type "string"}
+			;{type ["string" "null"]}
 			;the unique string identifier for the parent of the trainee
 			parent_id (null)
 			;{type "assoc" additional_indices "string"}
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -820,13 +820,13 @@
 			;the unique string identifier for the parent of the trainee
 			parent_id (null)
 			;{type "assoc" additional_indices "string"}
-			;map of contained trainee name to its unique id
+			;map of contained trainee path label to its unique id
 			path_to_id_map (null)
 			;{type "assoc" additional_indices "string"}
-			;map of contained trainee unique id to its name
+			;map of contained trainee unique id to its path label
 			id_to_path_map  (null)
 			;{type "assoc" additional_indices "boolean"}
-			;map of trainee name to boolean if it's contained internally
+			;map of trainee path label to boolean if it's contained internally
 			is_contained_map (null)
 		)
 		(call !ValidateParameters)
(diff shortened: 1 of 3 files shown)


howso/hierarchy.amlg
```diff
@@ -811,36 +811,78 @@
...
+			)
+		)
+
+		(if bump_revision (accum_to_entities (assoc !revision 1)) )
+
 		(call !Return)
 	)
 
+	;method to output references for contained trainees
+	;{idempotent (true) protected (true)}
```
fulpm: This getter should be marked `read_only` instead of `protected`

howso/hierarchy.amlg
```diff
@@ -811,36 +811,78 @@
...
-	#set_contained_trainee_maps
+	;method to update the references for contained trainees
+	;{idempotent (true) protected (true)}
+	#set_hierarchy_relationships
 	(declare
 		(assoc
+			;{type ["string" "null"]}
+			;the unique string identifier for the parent of the trainee
+			parent_id (null)
 			;{type "assoc" additional_indices "string"}
```
fulpm, line 822: Might also type these maps for "null" too, since it looks like they work the same as parent_id

howso/hierarchy.amlg
```diff
@@ -811,36 +811,78 @@
...
+		(if bump_revision (accum_to_entities (assoc !revision 1)) )
+
 		(call !Return)
 	)
 
+	;method to output references for contained trainees
+	;{read_only (true) idempotent (true)}
+	#get_hierarchy_relationships
+	(declare
+		;return {type "assoc"}
```
fulpm: Missing the return types for each property

howso/hierarchy.amlg
```diff
@@ -811,36 +811,101 @@
...
+	;{read_only (true) idempotent (true)}
+	#get_hierarchy_relationships
+	(declare
+		;returns {
+		; 	type "assoc"
+		;	additional_indices (false)
+		; 	indices {
+		; 		"parent_id" {
+		; 			type "string"
+		; 			description "The unique sting identifier for the parent of the tarinee."
```
fulpm:
```suggestion
		; 			description "The unique identifier for the parent of the trainee."
```

howso/hierarchy.amlg
```diff
@@ -811,36 +811,101 @@
...
 
-	;method to update the references for contained trainees, should only be used for debugging purposes.
-	;{idempotent (true)}
-	#set_contained_trainee_maps
+	;method to update the references for contained trainees
+	;{idempotent (true) protected (true)}
+	#set_hierarchy_relationships
 	(declare
 		(assoc
+			;{type ["string" "null"]}
```
cademack, line 819: This can just be "string" I believe

howso/return_typing.amlg
```diff
@@ -424,6 +424,10 @@
 									type "boolean"
 									description "Flag indicating if the label should be given the active session for the session parameter automatically from clients."
 								}
+								"protected" {
+									type "boolean"
+									description "Flag indicating this method should not be publicly available in the API."
```
cademack: the other description refer to the "label" rather than the method, could you change this to use "label"?

howso/hierarchy.amlg
fulpm: Its possible that we may allow moving a child trainee to be a new root, in which case we would want to set it to null

howso/hierarchy.amlg
```diff
@@ -811,36 +811,101 @@
...
+	;{read_only (true) idempotent (true)}
+	#get_hierarchy_relationships
+	(declare
+		;returns {
+		; 	type "assoc"
+		;	additional_indices (false)
+		; 	indices {
+		; 		"parent_id" {
+		; 			type "string"
+		; 			description "The unique string identifier for the parent of the tarinee."
```
fulpm: misspelled `trainee` here

howso/hierarchy.amlg
```diff
@@ -811,36 +811,101 @@
...
+		; 		"parent_id" {
+		; 			type "string"
+		; 			description "The unique string identifier for the parent of the trainee."
+		; 		}
+		; 		"path_to_id_map" {
+		; 			type "assoc"
+		;			additional_indices "string"
+		; 			description "Map of contained trainee path label to its unique id."
+		; 		}
+		; 		"id_to_path_map " {
```
fulpm: Excess space in parameter name

howso/hierarchy.amlg
```diff
@@ -811,36 +811,101 @@
...
+	;method to output references for contained trainees
+	;{read_only (true) idempotent (true)}
+	#get_hierarchy_relationships
+	(declare
+		;returns {
+		; 	type "assoc"
+		;	additional_indices (false)
+		; 	indices {
+		; 		"parent_id" {
+		; 			type "string"
```
fulpm:
```suggestion
		; 			type ["string" "null"]
```

fulpm (approved):

PR #363, opened 2024-11-21
22263: Updates hierarchy operations, MAJOR

howsoRes:
changes subtrainee methods to take in path and child_id and work directly instead of via execute_on_subtrainee()

### DIFF
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -391,8 +391,8 @@
 				(if (and (not separate_files) (= "amlg" !file_extension))
 					(store (concat filepath filename "." !file_extension) (flatten_entity trainee_path (false)) )
 
-					;store out the trainee and escape all filenames
-					(store_entity (concat filepath filename "." !file_extension) trainee_path (true) (true))
+					;store out the trainee as a caml
+					(store_entity (concat filepath filename "." !file_extension) trainee_path)
 				)
 		))
 
--- a/howso/hierarchy.amlg
+++ b/howso/hierarchy.amlg
@@ -198,14 +198,15 @@
 					child_entity_path trainee_path
 				))
 
-				(call !Return (assoc payload (assoc "path" result "id" child_id) ))
+				(call !Return (assoc
+					payload (assoc "path" (call !FriendlyPath (assoc path result)) "id" child_id)
+				))
 			)
 
 			;else error out
 			(call !Return (assoc errors (list "Failed to create trainee: ensure a valid filepath.")))
 		)
 	)
-
 
 	;Attempts to load a subtrainee with the following optional parameters.
 	;If the saved instance does not exist the existing trainee will remain unmodified and the function will return null.
@@ -277,7 +278,7 @@
 					is_contained (false)
 				))
 				(conclude
-					(call !Return (assoc payload (assoc "path" path) ))
+					(call !Return (assoc payload (assoc "path" (call !FriendlyPath)) ))
 				)
 			)
 		)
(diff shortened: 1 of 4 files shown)


howso/hierarchy.amlg
```diff
@@ -469,19 +413,54 @@
 	#delete_subtrainee
 	(declare
 		(assoc
-			;{type ["string" "list"] values "string" required (true)}
-			;name path of trainee
-			trainee (null)
+			;;;{type ["string" "list"] values "string" required (true)}
+			;;;name path of trainee
+			;;;trainee (null)
```
fulpm: Commented out typings? Should we just remove this

howso/hierarchy.amlg
```diff
@@ -356,14 +232,55 @@
 			;name to load (without extension)
 			filename ""
 			;{type ["string" "list"] values "string"}
```
fulpm: Seems like this type should just be "list"?

fulpm: Same goes for some of the other methods, all should accept the same type for path.

howso/hierarchy.amlg
```diff
@@ -356,14 +232,55 @@
...
-			;name path of trainee to load
-			trainee ""
+			;list of strings specifying the user-friendly path of the child subtrainee to load
+			;including the name of the child as the last value in the path
+			path (null)
+			;{type "string" required (true)}
+			;unique id of child trainee to load
+			child_id (null)
```
fulpm, line 250: if child_id is required, is there any need for path?

fulpm: Same goes for other non-create methods

howsoRes: both are required for load - path to specify where to load, and id so that the trainee_id is specified and can be used as the parent in case user decides to load subtrainees of this one

howso/hierarchy.amlg
```diff
@@ -506,17 +481,17 @@
 			target_trainee (null)
```
fulpm: Should target_trainee also just be a string?

howso/hierarchy.amlg
```diff
@@ -406,11 +323,17 @@
 				)
 		))
 
-		(if loaded_trainee
-			(call !Return (assoc payload (assoc "name" loaded_trainee) ))
-
-			(call !Return (assoc errors (list (concat "Failed to load trainee " filename))  ))
+		(if (= (null) loaded_trainee)
+			(conclude
+				(call !Return (assoc errors (list (concat "Failed to load trainee " filename))  ))
```
cademack: colon here?

cademack (approved):

unit_tests/ut_h_hierarchy_by_name.amlg
```diff
@@ -78,85 +38,88 @@
...
+				target_path ["A"]
+				target_id "uniqueid"
+				source_path ["A" "c"]
 			))
 	))
 	(print "Can't copy - duplicate child: ")
 	(call assert_same (assoc
 		obs (get result (list 1 "detail"))
-		exp "Specified target_trainee already exists."
+		exp "Specified trainee is already in the hierarchy."
```
fulpm: Maybe instead use `A trainee already exists in the hierarchy at path "...".` instead

howso/hierarchy.amlg
```diff
@@ -355,15 +241,56 @@
 			;{type "string"}
 			;name to load (without extension)
 			filename ""
```
fulpm: `filename` should be required and default to `(null)`

fulpm (approved):

cademack (approved):

PR #370, opened 2024-12-02
22317: Updates `#update_trainee` to support updating contained subtrainees, MAJOR

howsoRes:
Updates `#update_trainee` to account for subtrainees and upgrade them as well in a recursive fashion allowing for the entire horde of Trainees to be upgraded properly. 

Also added a call to `#!ValidateParameters` within `#get_num_training_cases` as it was missing there. This ensures that this method is not called with invalid parameters (any parameters in the case of this label). 

This PR also removes the inactive parameter, `separate_files` from `#upgrade_trainee`. Specifying this parameter after this change will return an error that the `separate_files` parameter is not supported for this label. Simply removing this parameter from the call should resolve the issue.

### DIFF
--- a/howso/get_cases.amlg
+++ b/howso/get_cases.amlg
@@ -89,6 +89,7 @@
 		; 	}
 		; }
 		(assoc)
+		(call !ValidateParameters)
 		(call !Return (assoc payload (assoc "count" (call !GetNumTrainingCases) ) ))
 	)
 
--- a/howso/upgrade.amlg
+++ b/howso/upgrade.amlg
@@ -214,6 +214,17 @@
 		)
 		(call !ValidateParameters)
 
+		(if (!= !revision 0)
+			(conclude
+				(call !Return (assoc errors
+					(list (concat
+						"This Trainee has had modifications and cannot be used for upgrading. "
+						"Please use a new Trainee to upgrade old Trainees."
+					))
+				))
+			)
+		)
+
 		;create a temporary subtrainee from the latest howso code
 		(declare (assoc
 			new_trainee_path
--- a/howso/upgrade.amlg
+++ b/howso/upgrade.amlg
@@ -426,7 +426,8 @@
 			case_and_session_entities
 		)
 
-		;TODO: migrate on contained trainees
+		(declare (assoc new_trainee_entity contained_entity_path))
+
 		;iterate over all the applicable migration script versions and apply each one in order
 		(call !execute_migration_scripts)
 
@@ -444,7 +445,7 @@
 
 					;recursively migrate all the attributes and contained entities
 					(call !MigrateSubtrainee (assoc
-						migration_trainee (current_value 1)
+						migration_trainee (append migration_trainee !traineeContainer (current_value 1))
 						contained_entity_path new_contained_entity_path
 						new_labels new_labels
 					))
@@ -465,7 +466,7 @@
 						(!= (null) (get (current_value) "auto_derive_on_train"))
 						(!= (null) (get (current_value) "derived_feature_code"))
 					))
-					(retrieve_from_entity "!featureAttributes")
+					(retrieve_from_entity new_trainee_entity "!featureAttributes")
 				)
 		)
 
@@ -485,7 +486,7 @@
 				)
 
 				;cache all the custom specified derivation code into trainee
-				(assign_to_entities (assoc
+				(assign_to_entities new_trainee_entity (assoc
 					!featureCustomDerivedMethods
 						(call !ComposeCustomDerivedMethods (assoc
 							custom_derived_features_map custom_derived_features_map
--- a/howso/upgrade.amlg
+++ b/howso/upgrade.amlg
@@ -476,7 +476,7 @@
 						(!= (null) (get (current_value) "auto_derive_on_train"))
 						(!= (null) (get (current_value) "derived_feature_code"))
 					))
-					(retrieve_from_entity new_trainee_entity "!featureAttributes")
+					(retrieve_from_entity "!featureAttributes")
 				)
 		)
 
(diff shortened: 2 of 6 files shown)


howso/upgrade.amlg
```diff
@@ -198,12 +198,16 @@
...
+			;	If specified, trainee_json_filepath is ignored
+			trainee_amlg_filepath (null)
+			;{type "string"}
 			;path from which to load previously exported meta.json and exp.json files.
 			;	If unspecified, expects them to be located in the base installation /migrations/ directory.
 			trainee_json_filepath (null)
 			;{type "boolean"}
-			;flag, if true will load each case from its individual file
-			separate_files (false)
+			;if true, then will import and use the trainee information in the subtrainee named ".old_trainee"
```
howsoRes: why does user need to know about a named subtrainee for the flow, this comment should be clearer, ie flag used by subtrainees when upgrading themselves automatically.  perhaps even rename the flag to match?

howso/upgrade.amlg
```diff
@@ -381,6 +396,75 @@
...
+					(lambda
+						(call_entity migration_trainee "debug_label" (assoc label (current_index 1) ))
+					)
+					(zip old_labels)
+				)
+		))
+		;assign all the old labels into the current entity/trainee
+		(assign_to_entities import_metadata_map)
+
+		(map
```
howsoRes, line 440: add comment that this moves cases and entities from old trainee into this new one

howso/upgrade.amlg
```diff
@@ -381,6 +396,75 @@
 		(call !Return)
 	)
 
+	#!UpgradeFromSubTrainee
```
howsoRes, line 413: will need a nice comment block about how this flow works from a high level on this method since it's recursive

howso/upgrade.amlg
```diff
@@ -235,27 +239,38 @@
...
-		;if not importing metadata from saved file, upgrade trainee in place
-		(if (= (null) trainee_json_filepath)
+		;if amalgam trainee is specified, load and upgrade it
+		(if (or trainee_amlg_filepath preloaded)
 			(let
 				(assoc
-					;create a list of all the matadata labels
-					labels_to_keep (call_entity new_trainee_path "get_export_attributes")
-					old_labels (call get_export_attributes)
+					;t (call (load (concat trainee_amlg_filepath trainee ".amlg") ))
```
howsoRes: get rid of commented out code

howso/upgrade.amlg
cademack: Only sorta recursive now... recursive across entities? I'll still comment lol

howso/upgrade.amlg
cademack: well users could use this if they for some reason wanted to load the trainee themself

howso/get_cases.amlg
```diff
@@ -89,6 +89,7 @@
 		; 	}
 		; }
 		(assoc)
+		(call !ValidateParameters)
```
jdbeel, line 92: Is this related to the PR title or just a fix?

howsoRes: and why does it need to validateparameters when this method has none?

howso/upgrade.amlg
```diff
@@ -235,27 +253,36 @@
...
-		;if not importing metadata from saved file, upgrade trainee in place
-		(if (= (null) trainee_json_filepath)
+		;if amalgam trainee is specified, load and upgrade it
+		(if (or trainee_amlg_filepath preloaded)
 			(let
 				(assoc
-					;create a list of all the matadata labels
-					labels_to_keep (call_entity new_trainee_path "get_export_attributes")
-					old_labels (call get_export_attributes)
+					t
```
howsoRes: rename to old_trainee ?

howso/upgrade.amlg
```diff
@@ -330,42 +357,42 @@
...
-				(lambda
-					(create_entities
-						(current_index)
-						(set_type
-							(zip_labels
-								(indices (current_value))
-								(values (current_value))
+				;overwrite this trainee code with the new code, this does not overwrite contained entities (cases, sessions),
+				;but does overwrite all labels including versions with new ones
+				(assign_entity_roots (retrieve_entity_root new_trainee_path))
```
howsoRes: is this step even necessary since the 'new trainee' is what's doing the upgrading and for subtrainees, those are also clones of new_trainee_path already (from the clone_entities line inside !UpgradeFromSubTrainee)?

unit_tests/ut_h_upgrade_htest.amlg
```diff
@@ -0,0 +1,215 @@
...
+	(call keep_result_payload)
+	(print "Num training cases in c2: ")
+	(call assert_same (assoc
+		obs (get result "count")
+		exp 2
+	))
+
+	(call exit_if_failures {msg "Cases not preserved correctly."})
+
+	(if (= (system "os") "Windows")
```
howsoRes, line 217: probably outside the scope the scope of this task, but we should add an optional parameter to 'exit_if_failures' in unit_test to call a cleanup method on exit when failed, then slap a cleanup label on this `if` statement, that way all the exit_if_failures in this test will still delete the htest.amlg file even if the test fails.

howso/get_cases.amlg
cademack: completely unrelated, its purpose here is to error out if unexpected parameters (any parameters in this case) are given

howso/upgrade.amlg
cademack: sure

unit_tests/ut_h_upgrade_htest.amlg
cademack: sure, done

howso/upgrade.amlg
cademack: yeah I think you're right

howso/get_cases.amlg
jdbeel: Can you add a line to the PR description to indicate that this fix was made?

cademack: done

cademack (approved):

PR #378, opened 2024-12-05
22357: WIP, mda as feature weight for targetless

howsoRes:


howso/analysis_weights.amlg
```diff
@@ -422,13 +433,12 @@
 							feature_residuals_map
 								(call !CalculateFeatureResiduals (assoc
 									features features
-									robust_residuals (true)
+									robust_residuals computing_mda
+									superfull (true)
```
cademack: having 2 binary flags `robust_residuals` and `superfull` will probably be pretty confusing. Maybe a string parameter would be better long term?

howso/feature_residuals.amlg
```diff
@@ -424,6 +430,30 @@
 			))
 		)
 
+		(if (and robust_residuals superfull (= (null) target_residual_feature))
```
cademack: this is a painful way to signal that MDA should be computed :/

howso/mda_weight.amlg
```diff
@@ -0,0 +1,436 @@
...
+							features
+						)
+					)
+				)
+			series_id_features (if !tsTimeFeature (get !tsModelFeaturesMap "series_id_features") )
+		))
+
+		;iterate over each case and accumulate residuals for all the feature(s)
+		#!AccumulateAccurateFeatureResiduals
+		(assign (assoc
```
cademack, line 32: To match the other FeatureResiduals methods, this code should support having a separate lists of `features`: the features to compute the residuals for, and `context_features`: the features to use as contexts (which can overlap and probably will usually overlap)

PR #390, opened 2025-01-03
22453: Adds "feature_deviations" react detail, MINOR

cademack:
Adds "feature_deviations" detail, which compute the feature deviations for the regional data surrounding a prediction, and adds it into the output object.


howso/details_residuals.amlg
```diff
@@ -87,6 +87,59 @@
 		)
 	)
 
+	;Helper method to compute and add to output local feature residuals and optionally prediction stats
+	#!ComputeReactFeatureDeviations
+	(declare
+		(assoc
+			relevant_features (append context_features action_features)
+			regional_model_cases_map
+				(compute_on_contained_entities (append
```
howsoRes: if all you need is a list of cases, use `contained_entities` instead, rename this variable to something like regional_case_ids and pass that in directly into case_ids

cademack: ah, good point

cademack: fixed

howsoRes (approved):

PR #392, opened 2025-01-08
22488: Fixes an issue in `#react_series` where the first row of a series could only generate null feature values as all data was disallowed from the query

cademack:
When doing series generation with a universal time feature, a situation could often arise where the first row of a series would generate many null values as the time value generated would be before that of any data in the model. In this case, the internal queries to populate other feature values would restrict the use of cases with time values larger than that of the context, which would be all the training data in this context.

Here lies a fix that puts a minimum bound on how restrictive this logic can be. Specifically, the model will now always have access to at least the K'th lowest time-value in the trained data.

### DIFF
--- a/howso/react.amlg
+++ b/howso/react.amlg
@@ -1381,5 +1381,33 @@
 				use_case_weights use_case_weights
 			))
 		)
+
+		;when doing time-series flows, the minimum time bound needs to be defined
+		(if (and
+				!tsTimeFeature
+				!tsTimeFeatureUniversal
+				(not (get !tsModelFeaturesMap "minimum_time_bound"))
+			)
+			#!UpdateMinimumTimeBound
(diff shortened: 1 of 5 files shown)


howso/react.amlg
```diff
@@ -1381,5 +1381,33 @@
...
 		)
+
+		;when doing time-series flows, the minimum time bound needs to be defined
+		(if (and
+				!tsTimeFeature
+				!tsTimeFeatureUniversal
+				(not (get !tsModelFeaturesMap "minimum_time_bound"))
+			)
+			#!UpdateMinimumTimeBound
+			;update "minimum_time_bound" which is the stored minimum time value that can be used as an upper bound
```
howsoRes, line 1391: comments should go above the label for consistency with all other labels and methods

howsoRes (approved):

PR #398, opened 2025-01-13
22539: Add support for continuous rebalance features, MINOR

howsoRes:
a) moved case weight rebalancing to after analyze
b) added distance-contribution based case weighting for continuous features. 
c) added support for mixed (both) nominal and continuous based features for both analyze and train flow


howso/analysis_weights.amlg
```diff
@@ -515,9 +515,94 @@
...
+
+				;set global continuous_rebalance_features variable to store these values during future train() calls
+				(assign_to_entities (assoc !continuousRebalanceFeatures continuous_rebalance_features))
+
+				;store the assoc of case_id -> weight
+				(assign (assoc
+					case_continuous_weights_map
+						(map
+							(lambda
+								;convert surprisal to probabilities, smaller DC -> smaller probabibility
```
cademack: probabibility

howso/analysis_weights.amlg
```diff
@@ -515,9 +515,94 @@
...
-			all_cases_map (zip (call !AllCases))
+			continuous_rebalance_features
+				(filter
+					(lambda (not (contains_index !nominalsMap (current_value))))
+					id_features
+				)
+			hyperparameter_map
+				(call !GetHyperparameters (assoc
+					context_features context_features
+					weight_feature ".case_weight"
```
cademack: this should probably be the "case_weight" parameter, and not the literal?

howso/analysis_weights.amlg
```diff
@@ -515,9 +515,94 @@
...
+				))
+			;assoc of case_id -> weight from continuous rebalance features
+			case_continuous_weights_map (null)
+		))
+
+		(if (size continuous_rebalance_features)
+			(seq
+				(assign (assoc
+					id_features
```
cademack: maybe a painful ask, but could this method name and inner variable names not strictly reference "ID features"? I think it's gonna be a bit misleading/confusing over time as we look into balancing out features that are not ID features traditionally

howso/train.amlg
```diff
@@ -653,16 +653,189 @@
...
+
+		;time feature exists, do ablation on time series by deriving one series at a time and then ablating those cases prior to training
+		!tsTimeFeature
+		(call !TrainTimeSeriesAblation)
+
+		;else ablating cases during training
+		(call !TrainCasesWithAblation)
+	)
+
+	;sub method to train cases without ablation, computing and storing an estimated case weight if rebalance features are specified
```
cademack, line 666: should be weight feature parameter?

cademack: err - I guess that's not a parameter to #train... never mind. But that is likely to be a problem someday if I had to guess

howso/train.amlg
```diff
@@ -653,16 +653,189 @@
...
+
+				;if there are nominal rebalance features, pull the corresponding nominal value weights
+				(if (> (size (get !savedAnalyzeParameterMap "rebalance_features")) (size continuous_rebalance_features))
+					(let
+						(assoc
+							nominal_rebalance_features_set
+								(remove
+									(zip (get !savedAnalyzeParameterMap "rebalance_features"))
+									continuous_rebalance_features
+								)
```
cademack: use the ablation case weight if available

unit_tests/ut_h_id_features.amlg
```diff
@@ -19,27 +19,27 @@
 			data
 				(list
 					;		"h" "w" "l" 	"size" 	"color"
-					(list 	1 	.7 	1 		"small" 	"red")
-					(list 	1.1 .6 	.7		"small" 	"red")
-					(list 	.6 	.4 	.5 		"small" 	"red")
-					(list 	.7 	.5 	.6 		"small" 	"green")
+					(list 	1 	.7 	1 		"small" 	"red"	1)
```
cademack, line 22: add `"popularity"` on line above?

howsoRes: no, it shouldn't be part of the regular features, only used in the later tests and will be explicitly referred to when needed

cademack (approved):

howso.amlg
```diff
@@ -698,6 +699,9 @@
 
 			;flag to indicate if the DD subtrainee should be recreated and used with each iteration of residuals within Analyze
 			!useDynamicDeviationsDuringAnalyze (false)
+
+			;list of rebalance features that are continuous, set if case weights are enabled and computed from these features
+			!continuousRebalanceFeatures (null)
```
howsohazard, line 705: Why do we need !continuousRebalanceFeatures and not one for nominal or continuous?  Just curious because it's not clear at first glance in the code documentation

cademack (approved):

howsohazard (approved):

howsohazard (approved):

PR #406, opened 2025-01-21
22584: Improves ablation check in !TrainCreateCases to prevent undesired warning

jdbeel:
Since `!TrainCreateCases` is called in `train` before `!AutoAnalyzeIfNeeded`, if auto-ablation and auto-analyze are enabled a warning can be raised regarding not having analyzed with case weights even if all parameters are set correctly.

This is not observed in larger use-cases like the Engine recipe or benchmarks because this is only relevant when the train batch size is $\geq$ both the auto-analyze threshold and the auto-ablation minimum model size.

### DIFF
--- a/unit_tests/ut_iris.amlg
+++ b/unit_tests/ut_iris.amlg
@@ -755,9 +755,13 @@
 		cases (list (list 2.2 1.5 0.1 "setosa" (- expected_value 0.8)))
 	))
 	(print "size " (call !get_num_cases) "\n")
-	;expected model size should now be 40
+	;expected model size should now be 40-42
 	(print "Model size after ablation training: " (call !get_num_cases) "\n")
-	(call assert_same (assoc obs (call !get_num_cases) exp 40))
+	(call assert_true
+		(assoc obs
+			(contains_value [40 41 42] (call !get_num_cases))
+		)
+	)
 
 	(call_entity "howso" "set_auto_ablation_params" (assoc
 		auto_ablation_enabled (false)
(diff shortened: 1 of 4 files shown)


cademack (approved):

unit_tests/ut_iris.amlg
```diff
@@ -755,9 +755,9 @@
 		cases (list (list 2.2 1.5 0.1 "setosa" (- expected_value 0.8)))
 	))
 	(print "size " (call !get_num_cases) "\n")
-	;expected model size should now be 42
+	;expected model size should now be 40
```
howsoRes: the train calls above on lines 748 and 753 should train those two cases, see the comments, so the result should be 42 as it was. why is this changed? is the expected value different now? should the values being passed in be different on lines 750 and 755?

unit_tests/ut_iris.amlg
```diff
@@ -755,9 +755,13 @@
 		cases (list (list 2.2 1.5 0.1 "setosa" (- expected_value 0.8)))
 	))
 	(print "size " (call !get_num_cases) "\n")
-	;expected model size should now be 42
+	;expected model size should now be 40-42
 	(print "Model size after ablation training: " (call !get_num_cases) "\n")
-	(call assert_same (assoc obs (call !get_num_cases) exp 42))
+	(call assert_true
+		(assoc obs
+			(contains_value [40 41 42] (call !get_num_cases))
```
howsoRes: the train calls above on lines 748 and 753 should train those two cases, see their corresponding comments, so the result should always be 42 as it was. why is this changed? why is the expected value different now? should the values being passed in be different on lines 750 and 755?

fulpm (approved):

cademack (approved):

howsoRes (approved):

PR #409, opened 2025-01-28
22631: Corrected ablated indices and session training indices when training with ablation, MAJOR

fulpm:


migrations/migrations.amlg
```diff
@@ -431,5 +431,25 @@
 			!hyperparameterMetadataMap new_hp_params
 		))
 	)
-
+"95.0.0"
```
fulpm, line 434: Need to test this migration

howso/train_ts_ablation.amlg
```diff
@@ -334,102 +325,197 @@
 						)
 					)
 				)
-			trained_series_cases []
+			trained_series_cases_features (append features [".series_index" ".case_id"])
+			trained_series_cases [] ;cases with the same series id that are already trained
```
cademack: comments should be above variables - unless we shifting styles. I'd prefer we keep them above, it's an Amalgam thing

howso/train_ts_ablation.amlg
```diff
@@ -334,102 +325,197 @@
...
+					case_id_feature_index (- (size trained_series_cases_features) 1) ;case_id index in trained_series_cases
 				)
 
-				;overwrite trained_series_cases to contain each cases's feature values and all the progress_features values
+				;set trained_series_cases to contain each cases's feature values, series index, and case id
 				(assign (assoc
 					trained_series_cases
 						(map
 							(lambda
+								;NOTE columns here must match trained_series_cases_features
```
cademack: colon after NOTE?

howso/train_ts_ablation.amlg
```diff
@@ -334,102 +325,204 @@
...
+											(assign (assoc prev_series_index 0))
+											(null) ;return null
+										)
 
-								)
-								0 series_continuation_index 1
+										;else if there is a gap since the last series index, output null
+										(> (- series_index prev_series_index) 1)
+										(seq
+											(accum (assoc prev_series_index 1))
```
cademack, line 442: Res/Chris might be able to provide further insight/clarification. But I am under the impression that merging these would be slightly more efficient than having a separate accum and assign.

With something like 
```
(assign (assoc
	prev_series_index series_index
	next_series_row (+ next_series_row 1)
))
```

howso/train_ts_ablation.amlg
```diff
@@ -334,102 +325,204 @@
...
+				;map of index -> original index
+				(assign (assoc
+					ts_ablated_indices_map
+						(zip
+							(indices data)
+							(map (lambda (last (current_value))) data)
 						)
 				))
 			)
 		)
```
cademack, line 517: I would guess that unzipping (null) would still return (null), so maybe the if isn't necessary here

fulpm: unzip? Not sure if you highlighted the right line here

fulpm: The unzip will return an empty list instead of a null, so added the check to explicitly return null for the ablated case.

cademack (approved):

howsoRes (approved):

PR #414, opened 2025-02-03
22734: Adds a warning to `react` and `single_react` to warn users if any derived action features are not derivable given other specified features

cademack:


unit_tests/ut_h_time_series_stock.amlg
```diff
@@ -671,6 +671,29 @@
...
+						[1000]
+					]
+				desired_conviction (null)
+			))
+	))
+
+	(call assert_true (assoc
+		obs
+			(contains_value
+				(get result [1 "warnings"])
```
howsoRes: do a (call keep_result_warnings) above the assert_true and simply check result here

howsoRes (approved):

PR #418, opened 2025-02-05
22755: Adds goal_features_map to react, MINOR

howsoRes:


howso/react.amlg
```diff
@@ -285,6 +285,14 @@
 			;	allow_null - default is true, if true nulls may be generated per their distribution in the data
 			; Only used when desired_conviction is specified
 			feature_bounds_map (assoc)
+			;{type "assoc" additional_indices {ref "GoalFeatures"}}
+			; assoc of :
+			;		{ feature : { "goal" : "min"/"max", "value" : somevalue }}
+			;		that specifies explicit goals for any features to attempt to reach via either 'goal' or a specific 'value'
```
howsohazard: What happens if both value and goal are specified for non-nominals?  Is that allowed? (probably not?)

howso/react_discriminative.amlg
```diff
@@ -20,6 +20,12 @@
 	;
 	; case_indices: optional pair (list) of session id and index, where index is the original 0-based session_training_index of the case as it was
 	;			trained into the session. If this case does not exist, outputs null.
+	; goal_features_map: optional assoc of:
+	;				{ feature : { "goal" : "min"/"max", "value" : somevalue }}
+	;				that specifies explicit goals for any features to attempt to reach via either 'goal' or a specific 'value'
+	;				"goal": "min" or "max", will make a prediction while minimizing or maximizing the value for the feature or
+	;				"value" : somevalue, will make a prediction while approaching the specified value
+	;			note: nominal features only support 'value', for non-nominals, if both are provided, only 'goal' is considered.
```
howsohazard: same per comment above

howso/synthesis.amlg
```diff
@@ -7,6 +7,12 @@
 	; context_values: optional, list of initial feature values that are used to condition the generated case
 	; action_features: full list of features to generate values for the case. Context feature values, if provided, will be output as-is
 	; use_regional_residuals: flag, if false uses global residuals, if true calculate and uses regional residuals. Default is true.
+	; goal_features_map: optional assoc of:
+	;				{ feature : { "goal" : "min"/"max", "value" : somevalue }}
+	;				that specifies explicit goals for any features to attempt to reach via either 'goal' or a specific 'value'
+	;				"goal": "min" or "max", will generate a case while minimizing or maximizng the value for the feature, or
+	;				"value" : somevalue, will generate a case while approaching the specified value
+	;			note: nominal features only support 'value', for non-nominals, if both are provided, only 'goal' is considered.
```
howsohazard: ditto

howso/react_utilities.amlg
```diff
@@ -878,4 +881,83 @@
 			derived_action_features
 		)
 	)
+
+	;Helper method for reacts to update contexts with goal feature values from results of local_model_cases_tuple
+	#!UpdateContextsForGoalFeatures
+	(assign (assoc
+		;set feature weight of goal features to be half
```
howsohazard: to be half.....?

howso/react_utilities.amlg
```diff
@@ -878,4 +881,83 @@
...
+		feature_weights
+			(if feature_weights
+				(let
+					(assoc total_weight (apply "+" (remove feature_weights goal_features)) )
+
+					(append
+						feature_weights
+						(zip
+							goal_features
+							(/ total_weight (size goal_features))
```
howsohazard: Shouldn't this be (/ 1 total_num_features_with_goal_features)? (however is fastest to compute total_num_features_with_goal_features here)

howso/react_utilities.amlg
```diff
@@ -878,4 +881,83 @@
...
+		;set goal to max/min/specified value in local space
+		context_values
+			(append
+				context_values
+				(map
+					(lambda (let
+						(assoc goal_feature (current_value 1) )
+
+						(if (= "max" (get goal_features_map [goal_feature "goal"]))
+							(apply "max"
```
howsohazard: This and min below might be faster by doing a query.

howso/react.amlg
```diff
@@ -285,6 +285,16 @@
 			;	allow_null - default is true, if true nulls may be generated per their distribution in the data
 			; Only used when desired_conviction is specified
 			feature_bounds_map (assoc)
+			;{type "assoc" additional_indices {ref "GoalFeatures"}}
+			; assoc of :
+			;		{ feature : { "goal" : "min"/"max", "value" : somevalue }}
+			;		defining goal features, forces reevaluation of local data in reacts to pull the predicted action values toward
+			;		achieving the specified value or goal as defined by this map. Valid keys in the map are:
+			;		"goal": "min" or "max", will make a prediction while minimizing or maximizing the value for the feature or
+			;		"value" : somevalue, will make a prediction while approaching the specified value
```
cademack: "a feature value"

howso/react_utilities.amlg
cademack: I feel like the strength of the goal-leaning tendency is probably best as an additional parameter

howso/react_utilities.amlg
```diff
@@ -878,4 +881,94 @@
...
+							goal_features
+							(/ total_weight (size goal_features))
+						)
+					)
+				)
+			)
+		context_features (append context_features goal_features)
+
+		;set goal to max/min/specified value in local space
+		context_values
```
cademack: updating the external context_values and context_features will affect all the later computations like react details. If we want this - then I suppose its okay, but otherwise this might need to be written in a way where the influence weights are updated with an additional query inside this method and its scope

howso/react_utilities.amlg
```diff
@@ -878,4 +881,94 @@
...
+								)
+							)
+						)
+					))
+					goal_features
+				)
+			)
+
+		;limit the secondary goal-oriented query to only run on cases in this local space
+		custom_extra_filtering_queries
```
cademack: having this (query_in_entity_list) added into the custom_extra_filtering_queries is especially troublesome if this variable is used in any later queries for details

howso/types.amlg
```diff
@@ -591,6 +591,30 @@
...
+					(assoc
+						goal
+							(assoc
+								type "string"
+								enum (list "min" "max" )
+								description "Type of goal."
+							)
+						value
+							(assoc
+								type ["number" "string"]
```
cademack: just go "any" here. We try not to assume what a feature value can be

howso/react.amlg
```diff
@@ -1339,11 +1339,16 @@
...
 		(if goal_features
-			(seq
+			(let
+				(assoc
+					orig_feature_weights feature_weights
+					orig_context_features context_features
+					orig_context_values context_values
+					orig_custom_extra_filtering_queries custom_extra_filtering_queries
+				)
 				(call !UpdateContextsForGoalFeatures)
```
cademack: Maybe make this !UpdateContextsForGoalFeatures method take an additional parameter that defines the name of the query label to call and have it return the updated local_model_cases_tuple so that we don't need to store the original params and reassign them later.

cademack (approved):

PR #424, opened 2025-02-11
22759: Updates `react_aggregate` to use derivation logic for derived features and adds other logic to improve aggregate details when using time-series data

cademack:
Updates `react_aggregate` to predict features using their derivation logic rather than directly interpolating the values of influential cases. For now this will be default behavior and will always apply to features where it can, but it is planned to be parametrizable in the future.

Additionally adds the time-series context filtering logic that has been present in residuals to MDA and contributions.


howso/contributions.amlg
```diff
@@ -20,6 +20,7 @@
 	;				  Defaults to 300 + 2 * (number of features).
 	; run_on_local_model: optional, if true will use all the provided cases and not store results to model
 	; context_condition_filter_query: a list of queries that can be used to filter the possible set of cases that can be used for predictions
+	; use_derivation_logic: flag, if true action feature will be derived and the features needed to derive will be predicted
```
jdbeel, line 23: Did I read something about us excluding this for now, or did I hallucinate that? I guess it doesn't have to be removed from here, just not exposed in engine-py.

howso/contributions.amlg
```diff
@@ -55,11 +56,81 @@
...
+
+		(if deriving_action_feature
+			(seq
+				(declare (assoc
+					features_for_derivation (indices (get_all_labels (parse (get !featureAttributes [action_feature "derived_feature_code"]) )))
+					;will be used to indicate which features to predict for derivation
+					features_to_predict []
+				))
+
+				(if !tsTimeFeature
```
jdbeel, line 84: I assume non-time-series derived actions are to-be-implemented later? Might be worth making a ticket for this and putting it in a TODO so it doesn't get lost.

howso/contributions.amlg
```diff
@@ -229,19 +326,16 @@
 												(zip (filter (lambda (not (get remaining_feature_flags (current_index)) )) context_features) )
 										))
 
+										(if deriving_action_feature
+											(assign (assoc
+												features_to_predict (filter (lambda (not (contains_value filtered_context_features (current_value)))) features_for_derivation)
```
jdbeel: I see this is reused once... Not strictly necessary, but it might be good to put this at a label so it can be edited in one place?

howso/contributions.amlg
jdbeel: Also this (and other API additions) mean the PR should be marked MINOR

cademack: yeah, keeping the param out of the external API temporarily. leaving a ticket aside to put it in the react_aggregate API formally with some deeper control for the user than I have implemented now

howso/contributions.amlg
cademack: non-time-series derived features should still work, but they will not have this "clever" logic on which feature needed for derivation will be held out automatically (bc how would we know)

cademack: which is why only that second bit is behind the if condition

howsoRes (approved):

howsoRes (approved):

PR #425, opened 2025-02-12
22760: Fixes rebalance weight to maintain total weight of data mass

howsoRes:


howso/analysis_weights.amlg
```diff
@@ -632,6 +632,27 @@
 					))
 					(zip nominal_rebalance_features)
 				)
+			num_cases (call !GetNumTrainingCases)
+		))
+
+		;update the weights by scaling each class's weight so that the total mass of the dataset remains the same, as though it was unweighted
```
howsohazard: Should do this in the loop above instead of having a second loop

howsohazard (approved):

PR #426, opened 2025-02-13
22834: Fixes encoded conditioned features 

howsoRes:
also fixes a few other misc small issues (resetting hyperparameters, properly passing in dt into goal features method)

### DIFF
--- a/howso/get_cases.amlg
+++ b/howso/get_cases.amlg
@@ -413,8 +413,7 @@
 			)
 		)
 
-
-		(if !hasStringOrdinals
+		(if (size !encodingNeededFeaturesSet)
 			(assign (assoc
 				condition
 					(map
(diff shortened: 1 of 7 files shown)


howso/get_cases.amlg
```diff
@@ -438,6 +437,20 @@
...
+								;leave string nominals as-is
+								(if (contains_index !stringNominalFeaturesSet (current_index))
+									(current_value)
+
+									(first
+										(call !ConvertFromInput (assoc
+											feature_values [(current_value 2)]
+											features [(current_index 2)]
+										))
+									)
```
cademack: I believe the (current_value 2) here, the value pointed to by the feature name can be a list or a single value.

cademack (approved):

howsohazard (approved):

PR #437, opened 2025-02-19
13226: Updates dynamic bandwidth selection to use built-in query capability

howsoRes:


howso/conviction.amlg
```diff
@@ -141,7 +143,7 @@
 		;closest k must be at least 2 smaller than model size, i.e., a model of 5 needs a K of 3 or less:
 		;when knocking out a case during conviction calculations, each remaining case searches for K cases around itself
 		;meaning that the K must be at least 2 less than the model size
-		(if (< model_size (+ closest_k 2))
+		(if (and (< model_size (+ closest_k 2)) (~ 0 closest_k))
```
cademack: should this just use your new k_parameter_value above?

howso/details_influences.amlg
```diff
@@ -36,7 +36,14 @@
 			(let
 				(assoc
 					;number of robust cases is max(100, K * ( ln(num_features) + 1) )
-					num_robust_cases (max 100 (* k_parameter (+ 1 (log (size (values features (true) )))) ) )
+					num_robust_cases
+						(max
+							100
+							(*
+								(if (~ 0 k_parameter) k_parameter 15)
```
cademack, line 43: Is the logic here intentionally different than what is written for ex post?

cademack (approved):

howso/conviction.amlg
howsoRes: no, because we only want to change the closest_k if the k parameter is not dynamic, so we still need that (~ 0 closest_k) check, and closest_k is a shorter variable name that happens to be the same exact value at this point if it isn't dynamic.

PR #444, opened 2025-02-25
22799: Adds new optional parameter to `#react_aggregate`, "features_to_derive", MINOR

cademack:
Adds a new parameter to `#react_aggregate`, "features_to_derive". This optional parameter allows users to specify a list of features whose values should be derived using logic defined in their feature attributes rather than their values being predicted using interpolation of the most influential cases.


howso/influences.amlg
```diff
@@ -218,7 +218,7 @@
 									(get_all_labels (parse (get !featureAttributes [(current_index 1) "derived_feature_code"]) ))
 								))
 							)
-							(zip (filter (lambda (not (contains_index !derivedFeaturesMap (current_value)))) action_features))
+							(filter (lambda (contains_value features_to_derive (current_index))) (zip action_features))
```
howsoRes: `(keep (zip action_features) feature_to_derive)` is simpler/faster

howsoRes (approved):

PR #445, opened 2025-02-26
22928: Refactors and fixes rebalance weights to work with ablation

howsoRes:
changes:
- reduce auto re-analyze growth amount from e^2 to e since analyze is much faster
- adds a new built in feature ".probability_mass" as !internalLabelProbabilityMass for rebalance flows to store the unweighted accumulated mass of a case. This value starts at 1.0 for all cases and accumulates based on the amount of mass received from ablated/removed cases.  
- adds a new dataset attribute !cachedTotalMass which is the total trained and accumulated mass of all cases ever trained (ablated or not) into the dataset. This way if 10 cases were trained and 10 were ablated, this !cachedTotalMass will be 20 and if the probability mass of the 10 remaining cases were added up it would also be 20
- adds a new dataset attribute !cachedRebalanceTotalMass which is the total of all rebalanced and un-scaled case_weights. e.g., for a dataset with a nominal rebalance feature with 2 classes, the unscaled rebalance weight for each case is the reciprocal of its class count, once added up will sum up to 2. additionally trained cases will increase this total (until the next analyze)
- distance contributions being used for continuous rebalance features are now scaled correctly by probability mass
- rebalanced case weights are scaled up by !cachedTotalMass / !cachedRebalanceTotalMass so that the total of all trained cases' case weights always adds up to !cachedTotalMass, (as it would if there was no rebalancing), regardless whether rebalance features are nominal, continuous or a mix.
- as cases are trained, they set their rebalance weights to an approximately correct value in an attempt to keep the dataset rebalanced (until the next analyze)
- for data reduction or ablation, as cases are ablated/removed and their probability mass is distributed among the neighbors, the neighbor's case weights are recomputed accordingly, that way when case weights are recomputed on analyze, any ablation-accumulated or distributed mass is not lost while maintaining rebalancing.


howso/analysis_weights.amlg
```diff
@@ -554,244 +555,207 @@
...
-				;store distance contributions into cases here if there are no nominal rebalance features
-				(if (= 0 (size nominal_rebalance_features))
-					(call !StoreCaseValues (assoc
-						label_name weight_feature
-						case_values_map case_continuous_weights_map
+				(if !autoAblationEnabled
+					;distance contributions are multiplied by their own case weights for anomaly detection flows
+					;but for rebalancing, the more dense a case, the less it should contribute/the smaller its influence should be
+					;thus the need to divide this scaled DC the square of its own probability mass,
+					;to revert back to the base DC value and to scale down the base DC by that mass
```
howsohazard: It's clear what you're doing, but not exactly why. Can you spell it out a little more why you need to divide it twice?

howso/analysis_weights.amlg
```diff
@@ -554,244 +555,207 @@
...
-			;as the product of all nominal rebalance feature case weights for each case
-			(call !StoreCaseValues (assoc
-				case_values_map
-					(map
-						(lambda (let
-							(assoc
-								case_id (current_index 1)
-								feature_values_map (current_value 1)
+		(declare (assoc
+			total_mass (if (not !autoAblationEnabled) (call !GetNumTrainingCases))
```
howsohazard: What happens if auto ablation is enabled?  This is just null?

howso/conviction.amlg
```diff
@@ -901,22 +909,66 @@
...
-		;if all the neighbors have zero distance, return a 0
 		(if (= "surprisal_to_prob" dt_parameter)
 			(let
 				(assoc
-					probabilities (map (lambda (exp (- (current_value)))) (values closest_cases_distances_map))
+					probabilities
+						;multiply probabilities by their corresponding case weight
+						(if dc_weight_feature
+							;ensure order of probabilities stays consistent by unzipping by indices of closest_cases_distances_map
+							(unzip
```
howsohazard, line 924: Can't just use amalgam for the surprisal to prob here?  It should work as expected for distance contribs

howsoRes: for unweighted, yes

howsohazard (approved):

PR #450, opened 2025-03-02
23054: Improves early exploration for RL flows

howsoRes:
This change increases the chance of using nominal feature domain probabilities instead of global ones for extremely small datasets (ie, at the start of RL flows)

### DIFF
--- a/howso/synthesis_utilities.amlg
+++ b/howso/synthesis_utilities.amlg
@@ -313,6 +313,8 @@
 						(indices global_class_probabilities_map)
 						(/ 1 (size global_class_probabilities_map))
 					)
+				;cap max desired_conviction at 1 to increase chance of using these domain probabilities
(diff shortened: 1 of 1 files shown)


howso/synthesis_utilities.amlg
```diff
@@ -313,6 +313,8 @@
 						(indices global_class_probabilities_map)
 						(/ 1 (size global_class_probabilities_map))
 					)
+				;cap max desired_conviction at 1 to increase chance of using these domain probabilities
+				desired_conviction (if (> desired_conviction 1) 1 desired_conviction)
```
howsohazard, line 317: This is just shaping the learning rate, which will be problematic if we have a nominal class with a large number of rare unique values (e.g., 500 cases trained and 5000 possible values, even though 80% of the data is one value).

howsohazard: I'd rather not merge this in

PR #452, opened 2025-03-03
22983: Adds support for goal features to `#react_aggregate`, MINOR

cademack:
Adds support for specfying goal features and their values to `#react_aggregate`, allowing users to evaluate the performance of their model when attempting to optimize or condition their predictions on certain feature values/maxima.

This all takes place through the "goal_features_map" which is added to `#react_aggregate` in these changes.
This parameter will be used for predictions for all react_aggregate details.

### DIFF
--- a/howso/contributions.amlg
+++ b/howso/contributions.amlg
@@ -21,6 +21,7 @@
 	; run_on_local_model: optional, if true will use all the provided cases and not store results to model
 	; context_condition_filter_query: a list of queries that can be used to filter the possible set of cases that can be used for predictions
 	; derive_action_feature: flag, if true action feature will be derived and the features needed to derive will be predicted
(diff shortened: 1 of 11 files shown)


howso/contributions.amlg
```diff
@@ -21,6 +21,7 @@
 	; run_on_local_model: optional, if true will use all the provided cases and not store results to model
 	; context_condition_filter_query: a list of queries that can be used to filter the possible set of cases that can be used for predictions
 	; derive_action_feature: flag, if true action feature will be derived and the features needed to derive will be predicted
+	; goal_features_map: assoc of goal feature stuff, to be passed directly into DiscriminativeReact
```
howsoRes: **stuff**  :-D   please copy the docstring from #!ReactDiscriminative  , here and in the other places where this is in the docstring

cademack: ah yeah, I meant to update this. Good eye

PR #455, opened 2025-03-06
23079: Improves equidistance check during synthesis validation

howsoRes:
does an additional query to capture any cases that are equidistant to the farthest case in the local space that may have been missed due to floating point (im)precision

### DIFF
--- a/howso/synthesis_validation.amlg
+++ b/howso/synthesis_validation.amlg
@@ -288,7 +288,7 @@
 										(query_within_generalized_distance
 											(+
 												;epsilon for defining whether two values are equal within acceptable precision: 2 * num features * DBL_EPSILON
-												(* 2 (size (if has_novel_substitions non_novel_context_features context_features)) 10e-16)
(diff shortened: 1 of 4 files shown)


howso/synthesis_validation.amlg
```diff
@@ -274,18 +274,57 @@
...
+											(query_not_in_entity_list (list ignore_case))
+											(list)
+										)
+										;find any cases that are within epsilon of the farthest case in the local space,
+										;ignoring the cases that have alraedy been found
+										(query_not_in_entity_list (first local_model_cases_tuple))
+										(query_within_generalized_distance
+											(+
+												;epsilon for defining whether two values are equal within acceptable precision: 2 * num features * DBL_EPSILON
+												(* 2 (size (if has_novel_substitions non_novel_context_features context_features)) 10e-16)
```
howsohazard: Use 2.22045e-16, 10e-16 is too small

cademack (approved):

howso/synthesis_validation.amlg
howsohazard: Actually use 2.220446049250313e-16

PR #459, opened 2025-03-07
23120: Adds usage of extra case for distance ratio validation

howsoRes:


howso/distances.amlg
```diff
@@ -103,9 +110,9 @@
 					(call !QueryLocalModelMinDistanceContribution (assoc
 						feature_labels context_features
 						entity_ids_to_compute
-							(if (size extra_equidistant_cases)
+							(if (extra_equidistant_cases
```
cademack: extra opening paren here?

cademack (approved):

PR #465, opened 2025-03-12
23166: Increases epsilon used for equidistance checks

howsoRes:


howso/distances.amlg
```diff
@@ -67,19 +67,19 @@
...
 			extra_equidistant_cases
 				;look for extra equidistance cases only if the distance between last two cases are the same (within epsilon)
-				(if (<
+				(if (<=
 						(-
 							(last (last local_model_cases_tuple))
 							(get (last local_model_cases_tuple) (- (size (last local_model_cases_tuple)) 2))
 						)
-						2.220446049250313e-16
+						epsilon
```
howsohazard: rename this to dbl_machine_epsilon or something similar as to not be confused with differential privacy epsilon

howso/distances.amlg
```diff
@@ -107,13 +107,14 @@
 						(if extra_equidistant_cases
 							(append (first local_model_cases_tuple) extra_equidistant_cases)
 
-							(if (<= model_size k_parameter)
+							(if (<= model_size (size (first local_model_cases_tuple)))
```
howsohazard, line 110: can we remove the word "model" form this variable please?  just local_cases_tuple

howso/distances.amlg
howsohazard: or if this is modified and not just the 2.2e-16, then whatever qualifier is appropriate

howso/synthesis_utilities.amlg
```diff
@@ -985,4 +985,36 @@
 		)
 	)
 
+
+	#!ComputeEpsilon
```
howsohazard: should rename to include what kind of epsilon as to not confuse with epsilon from differential privacy.  also should comment on what it does

howso/synthesis_utilities.amlg
```diff
@@ -985,4 +985,36 @@
...
+						(zip (append context_features action_features derived_context_features))
+						(indices generated_uniques_list_map)
+					)
+
+					(zip (append context_features action_features derived_context_features))
+				)
+		)
+
+		(if (= "surprisal_to_prob" (get hyperparam_map "dt") )
+			;surprisal space epsilon is dbl_epsilon * ( num_nominal_features + 2 * num_continuous_features + num_features - 1 )
```
howsohazard: should comment that this is due to the per-distance term additions and subtractions in the operations

howso/distances.amlg
howsoRes: yup, diff pr tho

howsohazard (approved):

PR #483, opened 2025-03-31
22983: Adds support for goal features and goal dependent features to react_aggregate, MINOR

cademack:
Adds two parameters to `react_aggregate` which allow users to specify goals that will bias the sampling of cases that will be used to compute the desired metrics.

`goal_features_map` takes a map of feature names to goals (just like the `react` equivalent). Instead of changing the logic of predictions though, these specified goals instead alter the behavior of how the cases are sampled to compute each metric. Cases are sampled in a way that will still randomly select from the trained data, but will still bias towards optimizing the defined goals for each sampled region of the data.

`goal_dependent_features` takes a list of features and can be used to specify features that will not be ignored while optimizing the goal in the sampling process within each subregion of the data.

### DIFF
--- a/howso/get_cases.amlg
+++ b/howso/get_cases.amlg
@@ -39,6 +39,10 @@
 	; num: number of samples to return
 	; rand_seed: optional random seed
 	; case_weight_feature: optional, if provided will use this feature to do case weighted sampling
+	; query_conditions: a set of queries that can limits the available cases for sampling
+	; goal_features_map: the standard goal features map. When specified, will compute an extra query for
+	;	each originally sampled case which will select it's nearest case that optimizes the goal. Leaving
+	;	the number of desired cases that maximizes the goal but still is sampled randomly from the model.
 	#!SampleCases
 	(declare
 		(assoc
@@ -46,6 +50,7 @@
 			rand_seed (null)
 			case_weight_feature (null)
 			query_conditions (list)
+			goal_features_map (assoc)
 		)
 		(if (!= (null) num)
 			(contained_entities (append
@@ -61,6 +66,46 @@
 				query_conditions
 				(query_sample (call !GetNumTrainingCases) case_weight_feature (rand))
 			))
+		)
+	)
+
+	;for each case id given, gets its influential cases and selects the case that optimizes the goal
+	;parameters:
+	; goal_features_map: the map of goal features and their specified goals (value/max/min)
+	; case_ids: the originally sampled case_ids
+	; query_conditions: a set of queries that can limits the available cases for selection
+	#!SampleCasesForGoals
+	(declare
+		(assoc
+			goal_features_map (assoc)
+			case_ids (list)
+			query_conditions (list)
+		)
+
+		(declare (assoc
+			hyperparam_map
+				(call !GetHyperparameters (assoc
+					feature ""
+					weight_feature (if valid_weight_feature weight_feature)
+				))
+		))
+
+		(map
+			(lambda
+				(let
+					(assoc
+						case_id (current_value 1)
+					)
+
+					(declare (assoc
+						nearest_case_ids
+							(compute_on_contained_entities )
+					))
+
+					;TODO
+				)
+			)
+			case_ids
 		)
 	)
 
--- a/howso/get_cases.amlg
+++ b/howso/get_cases.amlg
@@ -78,8 +78,10 @@
 	(declare
 		(assoc
 			goal_features_map (assoc)
+			goal_dependent_features (list)
 			case_ids (list)
 			query_conditions (list)
+			context_features (list)
 		)
 
 		(declare (assoc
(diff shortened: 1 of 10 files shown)


unit_tests/ut_h_goal_features_agg.amlg
```diff
@@ -0,0 +1,91 @@
...
+	;to evaluate what features are important to sell a house while maximizing price,
+	;the engine accurately reports that wine cellars and country clubs are more important.
+	;Conversely, when the goal features are not specified, we should see these features be
+	;not nearly as important (there are far less "upper"="yes" homes than not).
+	(declare (assoc
+		data (tail model_data)
+		features (first model_data)
+	))
+
+	(assign_to_entities "howso" (assoc trainee "model"))
```
howsoRes: can you remove this line? i don't see how it does anything

howso/react_utilities.amlg
```diff
@@ -924,80 +924,89 @@
...
-						goal_features
-					)
-				)
 
-			;limit the secondary goal-oriented query to only run on cases in this local space
-			custom_extra_filtering_queries
-				(if custom_extra_filtering_queries
-					(append custom_extra_filtering_queries (query_in_entity_list (first local_data_cases_tuple)) )
-					(query_in_entity_list (first local_data_cases_tuple))
+							(get context_map (current_value))
```
howsoRes, line 1003: add comment, something like:  ;else pull value from existing context_map

howso/get_cases.amlg
```diff
@@ -64,6 +65,127 @@
...
+		))
+
+		(map
+			(lambda
+				(let
+					(assoc
+						case_id (current_value 1)
+						context_map
+							(zip
+								(append features goal_dependent_features)
```
howsoRes: declare this append outside the map so it's a variable instead of having to append this static list every time

howso/get_cases.amlg
```diff
@@ -64,6 +65,127 @@
...
+								(retrieve_from_entity (current_value 1) (append features goal_dependent_features))
```
howsoRes: reuse that variable here too

howso/get_cases.amlg
```diff
@@ -64,6 +65,127 @@
...
+					weight_feature (if valid_weight_feature weight_feature)
+				))
+			goal_features (indices goal_features_map)
+			goal_dependent_features_set (zip goal_dependent_features)
+		))
+
+		;setting up variables to use goal_feature influence code correctly
+		(declare (assoc
+			feature_weights (get hyperparam_map "featureWeights")
+			context_features (indices (remove (zip !trainedFeatures) goal_features))
```
howsoRes: I think you can remove this context_features because you declare a fresh one for each case inside the map statement below, so it seems that this isn't used

howso/get_cases.amlg
```diff
@@ -64,6 +65,127 @@
...
+								)
+						))
+					)
+
+					(declare (assoc
+						context_values (unzip context_map context_features)
+					))
+
+					(declare (assoc
+						inf_cases_map
```
howsoRes: i'm not a fan of using 'inf' as a prefix, it's ambiguously similar to infinite/infinity ?  i suggest 'similar'

howsoRes: i'm not a fan of using 'inf' as a prefix, it's ambiguously similar to infinite/infinity ?  i suggest 'similar'

howso/get_cases.amlg
```diff
@@ -64,6 +65,127 @@
...
+		))
+
+		;setting up variables to use goal_feature influence code correctly
+		(declare (assoc
+			feature_weights (get hyperparam_map "featureWeights")
+			context_features (indices (remove (zip !trainedFeatures) goal_features))
+			dt_parameter (get hyperparam_map "dt")
+		))
+
+		(map
```
howsoRes: I think this method itself is not in a MT flow, right? because it's part of case sampling, therefore perhaps adding a || on it will improve performance since it does reacts in each iteration?

howso/get_cases.amlg
```diff
@@ -64,6 +65,127 @@
...
+					))
+
+					(assign (assoc
+						inf_cases_map
+							(call !UpdateLocalInfluencesForGoals (assoc
+								case_ids (append (indices inf_cases_map) case_id)
+								feature_weights feature_weights
+								context_features context_features
+								context_values context_values
+								custom_extra_filtering_queries (trunc custom_extra_filtering_queries)
```
howsoRes, line 178: good - you removed the "ignore the case itself " logic from this re-query, but please add a comment as to what this is for future sanity :)

howso/get_cases.amlg
cademack: yeah, believe you're right

unit_tests/ut_h_goal_features_agg.amlg
cademack: old copypasta I think

howsoRes (approved):

PR #485, opened 2025-04-09
23214: Refactor robust accuracy contributions, MAJOR

howsoRes:


unit_tests/ut_h_stats.amlg
```diff
@@ -671,7 +671,7 @@
 			(call_entity "howso" "react_aggregate" (assoc
 				action_feature "fruit"
 				;this is truly a trivial operation, so low samples is ok
-				num_robust_influence_samples 10
+				num_robust_prediction_contributions_samples 10
```
cademack: should specify both params here, since this computes accuracy contributions as well

howso/mda_weight.amlg
```diff
@@ -388,6 +393,28 @@
 			num_features_m_one (- (size features) 1)
 		))
 
+		(if output_raw_mda
+			(conclude
+				(zip
+					features
+					;transpose the matrix, so every feature has its own (column) MDA values
+					(map
+						(lambda (let
```
cademack, line 402: seems like one of the cases where using a let/assoc is a bit overkill. Switching to use current_index with a sufficient comment is good enough IMO

unit_tests/ut_h_stats.amlg
howsoRes: updated

howso/mda_weight.amlg
howsoRes: I'd rather keep it for code clarity (nested map confusion) since this is not a tight loop,

cademack (approved):

PR #491, opened 2025-04-11
23400: Update sampling robust residuals 

howsoRes:


howso/residuals.amlg
```diff
@@ -602,6 +618,25 @@
 				)
 			;create a backup of context_features for time series flows
 			original_context_features (if !tsTimeFeature (replace context_features))
+			num_features (size context_features)
+		))
+
+		;create a binomial probabilities map for enabling 0 - (num_features - 1) features
```
howsohazard: This is a description of what it's doing, but not why.  We need that, because this math isn't straightforward.

howsohazard (approved):

PR #493, opened 2025-04-14
23418: Significantly reduces memory usage during react_series

howsoRes:
React series would accumulate each row to 'series_data', every accumulation creating an in-memory copy, this is very inefficient as more rows are generated and series_data grows, so does the memory usage.
This PR limits 'series_data' to only be the current row plus the length of max lags needed for series generation.  I.E., if there is only one lag, series data is always kept to just two rows, the previous row and the current row being generated/derived.

The accumulation of the entire matrix is done inherently utilizing appending to (previous_result), which does not create an in-memory copy and drastically limits memory usage. This should notably improve performance and memory usage for larger
+longer series.

### DIFF
--- a/howso/react_series_utilities.amlg
+++ b/howso/react_series_utilities.amlg
@@ -1641,7 +1641,6 @@
 												(if num_existing_series_rows
 													(assign (assoc
 														total_progress initial_existing_progress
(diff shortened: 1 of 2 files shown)


howsoRes:
did not improve performance enough to warrant the amount of changes

PR #496, opened 2025-04-15
23325: Adds "series_residuals" detail to `react_series`, MINOR

cademack:
Adds the "series_residuals" boolean flag and "series_residuals_num_samples" parameter to the available details when using react series. Requesting series uncertainty returns a list of dictionaries indicating estimated uncertainties for each continuous derived feature (other than time) at each timestep of the returned series.

"series_residuals_num_samples" allows users to specify the number of generative series to use in estimating this uncertainty.

### DIFF
--- a/howso/react_series.amlg
+++ b/howso/react_series.amlg
@@ -784,8 +784,8 @@
 			prev_index (null)
 		))
 
-		;this infinite loop assumes time_values are sorted
-		;it should always find a time_value greater than time
+		;this "infinite" loop iterates using the incrementing (current_index) and assumes
+		;time_values is sorted. it should always find a time_value greater than time
 		;because of the checks above
 		(while (true)
 			(if (> (get time_values (current_index)) time)
--- a/howso/react_series.amlg
+++ b/howso/react_series.amlg
@@ -781,21 +781,28 @@
 		)
 
 		(declare (assoc
-			prev_index
-				;this infinite loop assumes time_values are sorted
-				;it should always find a time_value greater than time
-				;because of the checks above
-				(while (true)
-					(if (>= (get time_values (current_index)) time)
-						(conclude (- (current_index) 1))
-
-						(> (current_index) (size time_values))
-						;this should never be reached unless time_values are not sorted
-						;and no element is greater than time
-						(conclude (null))
-					)
-				)
+			prev_index (null)
 		))
+
+		;this infinite loop assumes time_values are sorted
+		;it should always find a time_value greater than time
+		;because of the checks above
+		(while (true)
+			(if (> (get time_values (current_index)) time)
+				(conclude
+					(assign (assoc prev_index (- (current_index) 1) ))
+				)
+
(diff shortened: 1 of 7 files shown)


howso/react_series.amlg
```diff
@@ -754,4 +754,64 @@
...
+			pre_domain_value (null)
+			;value to return if 'time' is greater than the last value of 'time_values'
+			post_domain_value (null)
+		)
+
+		(if (< time (first time_values))
+			(conclude pre_domain_value)
+		)
+
+		(if (> time (last time_values))
```
howsoRes: this could just be the 'elseif' of the first if statement above instead of its own separate if statement

howso/react_series.amlg
```diff
@@ -754,4 +754,64 @@
...
+		)
+
+		(declare (assoc
+			prev_index
+				(while (true)
+					(if (>= (get time_values (current_index)) time)
+						(conclude (- (current_index) 1))
+
+						(> (current_index) 1000)
+						(conclude "heh?")
```
howsoRes: they won't expect this value coming out, that's for sure. 

you shouldn't even need this check here anyway, just add a comment that a forever loop isn't possible because the last value of time_values is checked by the if statement above that outputs post_domain_value, so if it made it this far, at least one value in time_values will be >= time.

cademack: agreed, gonna still leave a more proper fall-out in case time_values end up not sorted somehow

howso/react_series_utilities.amlg
```diff
@@ -1680,4 +1688,169 @@
...
+
+	;helper method for ComputeSeriesUncertainty to compute the MAD for a collection
+	;of continuous values
+	#!ComputeMAD
+	(declare
+		(assoc
+			;the list of values to compute the MAD for
+			vals []
+		)
+
```
howsoRes:
```
        ;can't compute M.A.D. with 0 or 1 values
		(if (<= (size vals) 1)
		    (conclude 0)
		)
```

howsoRes (approved):

howso/react_series.amlg
```diff
@@ -754,4 +754,75 @@
...
+			(conclude post_domain_value)
+		)
+
+		(declare (assoc
+			prev_index (null)
+		))
+
+		;this infinite loop assumes time_values are sorted
+		;it should always find a time_value greater than time
+		;because of the checks above
```
howsohazard, line 789: comment that it's walking over all of the indices

howso/react_utilities.amlg
```diff
@@ -1010,4 +1010,33 @@
 		;call the react query again with the goal-oriented contexts and weights
 		(call (retrieve_from_entity query_label))
 	)
+
+	;helper method to compute the MAD for a collection of continuous values
```
howsohazard, line 1014: "mean absolute deviation (MAD)"

howsohazard (approved):

PR #500, opened 2025-04-18
23485: Speeds up full dataset similarity conviction

howsoRes:
when computing on the entire dataset, reuses the already-computed distance contributions that are stored into case_to_dc_map instead of computing each case's individual DC value

### DIFF
--- a/howso/conviction.amlg
+++ b/howso/conviction.amlg
@@ -438,7 +438,7 @@
 									feature_values (unzip (current_value 1) features)
 									use_case_weights use_case_weights
 									weight_feature weight_feature
-									ignore_case_id (current_index 1)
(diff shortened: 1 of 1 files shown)


howso/conviction.amlg
```diff
@@ -775,6 +775,7 @@
 	; features: list of features
 	; feature_values: list of corresponding feature values
 	; filtering_queries: (optional) list of filtering queries (such as ignoring a specific case) during computation
+	; ignore_case_id: optional, id of case to compute similarity conviction for
```
howsohazard: why call this ignore_case_id?  Shouldn't it just be "case_id"?

howsohazard (approved):

PR #506, opened 2025-04-22
23532: Fixes internal performance tests

howsoRes:


cademack (approved):

performance_tests/adult_test.amlg
```diff
@@ -104,7 +100,7 @@
 
 	(call_entity "howso" "react_aggregate" (assoc
 		context_features features
-		num_samples 1000
+		num_robust_residual_samples 1000
```
howsohazard: Why compute robust residuals here?  Why not regular?  We seem to be testing this in all the performance tests.

performance_tests/adult_test.amlg
```diff
@@ -23,10 +23,8 @@
 		submodel_size (null) ;10000
 		;set to true to also run analyze
 		do_analyze (true)
```
howsohazard: get rid of do_analyze in all but mnist

PR #515, opened 2025-04-29
23586: Fixes continuous dependency features with null_is_dependent

howsoRes:


cademack (approved):

mkoistinen (commented):
LGTM, but I wish there were unit tests for the case where the dependent feature(s) were nominal. This only tests continuous.

howsohazard (approved):

PR #522, opened 2025-05-08
23597: Adds `similarity_conviction` to `#react_group`, MINOR

cademack:
Adds `similarity_conviction` to `#react_group`, allowing users to compute the mean similarity conviction of the cases in each group.

I also move all the methods that are only used in `#react_group` into "react_group.amlg" (quite reasonably if you ask me).


howso/react_group.amlg
```diff
@@ -348,9 +366,267 @@
...
+			kl_divergence_removal_value (/ new_case_entropies_value (size case_ids))
+		))
+	)
+
+	;helper method for react group to compute the mean similarity conviction of a group of cases
+	#!ComputeGroupMeanSimilarityConviction
+	(seq
+		(declare (assoc
+			case_similarity_convictions
+				(map
```
howsoRes: can you do some performance testing with this? also i suggest adding a || on this map so it's parallized

cademack: added the pipe

howsoRes (approved):

PR #526, opened 2025-05-14
23723: Fixes issues that arise while using encoded features when computing forecast accuracy

cademack:
When using encoded features and computing forecast accuracy, the values compared were not comparable leading to misleading or invalid results.

also fixes a datetime conversion bug, and adds some more comments to RunForecastResiduals

### DIFF
--- a/howso/feature_residuals.amlg
+++ b/howso/feature_residuals.amlg
@@ -1457,6 +1457,7 @@
 
 		(declare (assoc
 			feature_index_map (zip all_features (indices all_features))
+			;parameters needed within ReactSeries
 			derived_context_features
 				(filter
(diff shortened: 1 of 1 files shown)


howsohazard (approved):

howso/feature_residuals.amlg
```diff
@@ -1508,6 +1510,8 @@
 			case_residuals_lists
 				||(map
 					(lambda
+						;For each case, forecast the til the window-length is elapsed and interpolate both
```
howsoRes, line 1513: until?

howsoRes (approved):

howso/feature_residuals.amlg
howsoRes: will fix in a followup pr

PR #527, opened 2025-05-15
23653: Implements ignoring inactive features in all flows

howsoRes:
This explicitly removes inactive features from the context_features when computing residuals, contributions, deviations, stats, etc, to increase performance, while also ensuring that their weights are zero'd out everywhere feature weights could be used (both featureWeights and featureMdaMap)

### DIFF
--- a/howso/contributions.amlg
+++ b/howso/contributions.amlg
@@ -115,7 +115,7 @@
 					excluded_inactive_features_map (keep !inactiveFeaturesMap context_features)
 				))
 				(assign (assoc
-					context_features (filter (lambda (not (contains_index !inactiveFeaturesMap (current_value)))) context_features)
+					context_features (filter (lambda (not (contains_index excluded_inactive_features_map (current_value)))) context_features)
 				))
 			)
 		)
--- a/howso/contributions.amlg
+++ b/howso/contributions.amlg
@@ -108,6 +108,7 @@
 			)
 		)
 
+		;don't use inactive features for computing prediction contributions
 		(if (size !inactiveFeaturesMap)
 			(seq
 				(assign (assoc
@@ -216,6 +217,7 @@
 			(call !RunFullContributions)
 		)
 
+		;put the inactive features back into output with 0s
 		(if (size excluded_inactive_features_map)
 			(accum (assoc feature_contributions_map excluded_inactive_features_map))
 		)
--- a/howso/contributions.amlg
+++ b/howso/contributions.amlg
@@ -62,6 +62,7 @@
 			;store an assoc of lag/rate/delta feature -> lag/order amount for time series flows
 			ts_feature_lag_amount_map (if !tsTimeFeature (call !BuildTSFeatureLagAmountMap))
 			max_lag_index_value (null)
+			excluded_inactive_features_map {}
 		))
 		(if ts_feature_lag_amount_map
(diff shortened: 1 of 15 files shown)


howso/contributions.amlg
```diff
@@ -107,6 +108,18 @@
...
 		)
 
+		;don't use inactive features for computing prediction contributions
+		(if (size !inactiveFeaturesMap)
+			(seq
+				(assign (assoc
+					excluded_inactive_features_map (keep !inactiveFeaturesMap context_features)
+				))
+				(assign (assoc
+					context_features (filter (lambda (not (contains_index !inactiveFeaturesMap (current_value)))) context_features)
```
cademack: any reason to (contains_index) !inactiveFeaturesMap rather than excluded_inactive_features_map here? I suppose it shouldn't matter too much. Just seems odd you make the more localized map just to use the global one again

howso/residuals.amlg
```diff
@@ -186,8 +186,12 @@
 			temp_residual_map
 				(map
 					(lambda
-						;set upper bound to nominals to be max nominal deviation, accounting for imbalanced classes
-						(if (contains_index !nominalsMap (current_index))
+						;inactive features don't expact residuals because they are all 0
```
cademack: expand*

howso/train_utilities.amlg
```diff
@@ -65,22 +65,44 @@
...
-													(assign (assoc
-														weights_map
-															(set
-																weights_map
-																(current_index 1)
-																(current_value 1)
+											(assign (assoc weights_map (append weights_map features_weights_map) ))
+
+											;if there is a featureMdaMap specified, temporarily set the weight of activated features to be 1/num_features
+											;so that they can be used in queries until the dataset is reanalized
```
cademack: re-analyzed*

howso/train_utilities.amlg
```diff
@@ -152,40 +174,48 @@
...
+								])
 							)
-							0
 						)
 					))
 					(if !inactiveFeaturesMap
 						!inactiveFeaturesMap
-						(zip !trainedFeatures 0)
+						(zip !trainedFeatures (range 0 1 (size !trainedFeatures) 1))
 					)
```
cademack: Why should this not always be over trainedFeatures?

This seems like it should always be over trainedFeatures bc if a user trains a new feature that needs to be selected as inactive, this will miss it?

howso/train_utilities.amlg
```diff
@@ -152,40 +174,48 @@
...
 				)
 		))
 
 		;if there are now less inactive features than there were, set the no-longer inactive feature weights to 1
 		(if (< (size inactive_features_map) (size !inactiveFeaturesMap))
```
cademack: And if you find my comment on the above correct, just checking the size of these is probably a bit irresponsible, but we should actually check the elements

howso/train_utilities.amlg
```diff
@@ -149,43 +171,61 @@
...
-								))
+									(query_not_equals feature feature_value)
+								])
 							)
-							0
 						)
 					))
-					(if !inactiveFeaturesMap
+					;once !inactiveFeaturesMap is created, add_feature and remove_feature edit it, so unless
+					;cases are beind edited, only need to check !inactiveFeaturesMap instead of all the trained features
```
cademack: being*

cademack (approved):

PR #541, opened 2025-05-29
23821: Improves the performance of analysis in datasets with many features that contain nulls

cademack:


howso/mda_weight.amlg
```diff
@@ -181,38 +181,42 @@
...
-														candidate_cases_lists
-															;only remove cases with null values if there are going to be some remaining
-															(if (< (size null_indices) (size (first candidate_cases_lists)))
 																(list
-																	(remove (first candidate_cases_lists) null_indices)
-																	(remove (get candidate_cases_lists 1) null_indices)
-																	(remove (last candidate_cases_lists) null_indices)
+																	(indices local_cases_map)
+																	(values local_cases_map)
+
```
howsoRes: get rid of this whitespace and the one below so this triple is just three lists in a row

howso/mda_weight.amlg
```diff
@@ -181,38 +181,42 @@
...
+															(assoc
+																filtered_local_cases_maps
+																	(compute_on_contained_entities [
+																		(query_in_entity_list (indices local_cases_map))
+																		(query_not_equals feature (null))
+																		(query_exists feature)
+																	])
+															)
+
+															(if (size filtered_local_cases_maps)
```
howsoRes, line 232: The original comment `;only remove cases with null values if there are going to be some remaining`
should be factored to back here, though it doesn't exactly apply anymore.  Thinking about it, the idea was that if there were any non-nulls, we wanted to interpolate between only those.  However, if they are all null, it would just interpolate among the original tuple and result in a null. Now we are always running this query, perhaps we can short-circuit that code that creates the full tuple in this 'all nulls' flow. 

I suggest either a) if (size filtered_local_cases_maps) is 0,  we can manually put in K (nulls)'s on line 206, or b) we can create the tuple with a single value something like `[ [(first (indices local_cases_map)] [1] [(null)] ]` since it'll be interpolated to a null anyway, why bother returning all k values when returning 1 will have the same result?

cademack: Option b) makes sense to me, that is what I will go with

howsoRes (approved):

howso/residuals.amlg
```diff
@@ -801,38 +801,41 @@
...
 																)
 															)
-													))
+														)
+
+														(list
+															(indices local_cases_map)
+															(values local_cases_map)
+
+															(map (lambda (retrieve_from_entity (current_value) feature)) (indices local_cases_map))
```
howsohazard, line 840: Should this be a query as well?  Or should I make a specialization of map that looks for this idiom?

howsohazard (approved):

PR #559, opened 2025-06-23
23956: Improves interpolation performance by avoiding accumulations

cademack:


howso/react.amlg
```diff
@@ -1508,20 +1508,19 @@
...
-								"class_counts"
-							])
-					))
-				)
-
-				exp_value
+				{
+					"action_value" exp_value
+					"categorical_action_probabilities_map"
+						(call !AccumulateCategoricalActionProbabilties (assoc
```
howsoRes: this should be in an if statement so that it doesn't do the call everytime

PR #561, opened 2025-06-25
23971: Checks that referenced entities exist in `#process_train_payload`

dmaze:
This is used when auto-ablation is on.  It's possible for one actor
to call compute_train_payload while another is executing a train-type
operation concurrently, which triggers automatic data reduction, which
causes child entities to be deleted.  When we do get to the "process"
step, double-check that all of the child entities we'd like to add
weights to are still there.

### DIFF
--- a/howso/scale.amlg
+++ b/howso/scale.amlg
@@ -332,16 +332,23 @@
 		)
 
 		;check that all of the referenced child entities still exist, and weren't deleted by another caller in between compute_train_payload and now
-		(declare (assoc referenced_entities (assoc)))
-		(if (size weight_accumulation_maps)
-			(assign (assoc referenced_entities (append referenced_entities (reduce (lambda (append (previous_result) (current_value))) weight_accumulation_maps))))
-		)
-		(if case_edit_map
-			(assign (assoc referenced_entities (append referenced_entities case_edit_map)))
-		)
-		(declare (assoc referenced_entity_names (sort (indices referenced_entities))))
-		(declare (assoc child_entity_names (sort (contained_entities (list (query_in_entity_list referenced_entity_names))))))
-		(if (!= referenced_entity_names child_entity_names)
+		(declare (assoc
+			referenced_entities (append
+				(if (size weight_accumulation_maps)
+					(reduce (lambda (append (previous_result) (current_value))) weight_accumulation_maps)
+					(assoc)
+				)
+				(if (size case_edit_map)
+					case_edit_map
+					(assoc)
+				)
+			)
+		))
+		(if
+			(<
+				(size (contained_entities (list (query_in_entity_list (indices referenced_entities)))))
+				(size referenced_entities)
+			)
 			(conclude (call !Return (assoc
 				errors ["Missing related training case"]
 				error_code "conflict"
--- a/howso/scale.amlg
+++ b/howso/scale.amlg
@@ -326,9 +326,26 @@
 		(call !ValidateParameters)
 
 		(if (not !autoAblationEnabled)
-			(call !Return (assoc
+			(conclude (call !Return (assoc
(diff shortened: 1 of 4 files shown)


howso/scale.amlg
```diff
@@ -326,9 +326,26 @@
...
+		(declare (assoc referenced_entities (assoc)))
+		(if (size weight_accumulation_maps)
+			(assign (assoc referenced_entities (append referenced_entities (reduce (lambda (append (previous_result) (current_value))) weight_accumulation_maps))))
+		)
+		(if case_edit_map
+			(assign (assoc referenced_entities (append referenced_entities case_edit_map)))
+		)
+		(declare (assoc referenced_entity_names (sort (indices referenced_entities))))
+		(declare (assoc child_entity_names (sort (contained_entities (list (query_in_entity_list referenced_entity_names))))))
+		(if (!= referenced_entity_names child_entity_names)
```
cademack: rather than having to sort these both and check their equality, could we instead just check if the size of `child_entity_names` is less than the size of `referenced_entity_names`?

cademack: or as an extra option, if you get them both as "sets" (assocs with null values), we could check if those are equal instead

howso/scale.amlg
```diff
@@ -326,9 +326,26 @@
...
+			(conclude (call !Return (assoc
 				errors ["`process_train_payload` should only be used on a Trainee that has auto-ablation enabled."]
-			))
+			)))
+		)
+
+		;check that all of the referenced child entities still exist, and weren't deleted by another caller in between compute_train_payload and now
+		(declare (assoc referenced_entities (assoc)))
+		(if (size weight_accumulation_maps)
+			(assign (assoc referenced_entities (append referenced_entities (reduce (lambda (append (previous_result) (current_value))) weight_accumulation_maps))))
```
cademack: This could probably be all built at once, with something like:

```
(declare (assoc
			referenced_entites
				(append
					(if (size weight_accumulation_maps)
						(reduce ... )
						(assoc)
					)
					(if (size case_edit_map)
						case_edit_map
						(assoc)
					)
				)
		))
```

which would be a bit more amalgam-ish IMO

unit_tests/ut_h_scale_ablation.amlg
```diff
@@ -0,0 +1,119 @@
...
+        session "unit_test"
+    ))
+
+    (print "analyze\n")
+    (call_entity "howso" "analyze" (assoc))
+
+    (print "reduce_data\n")
+    (call_entity "howso" "reduce_data" (assoc))
+
+    ; At this point, ablation has gotten rid of all of the cases between x=0 and x=300!
```
cademack, line 55: Nothing to say review-wise here, but this is particularly interesting if this is the true consistent behavior

dmaze: In my notebook flavor of it, it was nothing between 0 and 400.  And if you think about it, if the difference between 998^2 and 999^2 is about 2,000, the difference between 0^2 and 40^2 is comparatively less, you're not completely off to say that 5^2 is more or less zero on that scale.

The fact that there's such a huge gap here is a little suspicious to me, and I did want to comment in the code that this test passing depends on this specific ablation behavior, which could very reasonably change.

howso/scale.amlg
dmaze: I tried cleaning this up some, to avoid keeping the lists of names around.  Comparing the collection sizes is sound.  I like the option of comparing them as "sets" but I haven't been having a good time with Amalgam-flavored `map` today.

howso/scale.amlg
dmaze: Done.

cademack (approved):

PR #564, opened 2025-06-26
23987: Fixes an issue where no boundary cases could be found in time-series Trainees

cademack:
Fixes the issue that was caused by a few distinct issues. Here are the list of effective changes in this PR:

- "Un-shares" null uncertainties from shared deviations, these are now not shared and computed/stored in feature_deviations on a per-feature basis (regardless of if they are shared deviations features or not)
- Ignores the time-series filtering query within the boundary cases search logic
- Computes null uncertainties/deviations for *all features* even timeseries features whose nulls only exist when they push back before a series' start

### DIFF
--- a/howso/details_cases.amlg
+++ b/howso/details_cases.amlg
@@ -389,7 +389,7 @@
 					1000
 				)
 			;make a new list of filtering queries because time-series trainees will attempt to filter the future, not appropriate
-			;for time-serie
+			;for time-series
 			boundary_filtering_queries
 				(append
 					(if ignore_case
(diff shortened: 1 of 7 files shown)


howso/details_cases.amlg
```diff
@@ -388,6 +388,22 @@
 					(> specified_num_exemplars 1000)
 					1000
 				)
+			;make a new list of filtering queries because time-series trainees will attempt to filter the future, not appropriate
+			;for time-serie
```
howsoRes: missing s

PR #570, opened 2025-07-14
24057: Updates the math for the "feature_full_residual_convictions_for_case" detail to smooth out extremes

cademack:
Rather than divide the regional residual by the case residual directly. We now apply the LK metric to both residuals first, then divide. This makes the extreme conviction values returned when either residual is zero (or lesser than the deviation) much more reasonable to the user and prevents random extremes that result when dealing with very predictable regions of the data.

### DIFF
--- a/howso/details_residuals.amlg
+++ b/howso/details_residuals.amlg
@@ -562,19 +562,11 @@
 							(/
 								(call !LK_metric (assoc
 									difference (first (current_value 1))
-									deviation
-										(if (size global_residuals_map)
-											(get global_residuals_map (current_index 1))
-											0
-										)
+									deviation (or (get global_residuals_map (current_index 1)) 0)
 								))
 								(call !LK_metric (assoc
 									difference (get !cachedFeatureMinResidualMap (current_index 1))
-									deviation
-										(if (size global_residuals_map)
-											(get global_residuals_map (current_index 1))
-											0
-										)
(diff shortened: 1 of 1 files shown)


howso/details_residuals.amlg
```diff
@@ -598,4 +633,33 @@
...
 		))
 	)
+
+	;compute the lk given a distance (d) and a deviation (b)
+	;d should be positive already
+	#!LK_metric
+	(declare
+		(assoc
+			d (null)
+			b (null)
```
howsohazard: Probably just call difference and deviation to make it more clear

howsohazard (approved):

howso/details_residuals.amlg
```diff
@@ -598,4 +633,33 @@
...
+		)
+
+		(+
```
howsoRes: not sure you need to have this in a declare block.  can the !LK_metric just be directly on the code itself, so you don't need to declare d and b since they'll always be specified anyway?

howso/details_residuals.amlg
```diff
@@ -598,4 +633,33 @@
 				)
 		))
 	)
+
+	;compute the lk given a distance (d) and a deviation (b)
```
howsoRes: comment needs updating

howsoRes (approved):

PR #574, opened 2025-07-17
24086: Updates feature weighting logic to use the LK metric for exponential distributions

cademack:


howso/mda_weight.amlg
```diff
@@ -516,28 +516,37 @@
...
 										lk_value
 											(+
 												mda
 												(*
-													0.5
-													(exp (- (/ mda target_feature_residual)) )
-													(+ (* 3 target_feature_residual) mda)
+													2
+													(/
+														(pow target_feature_residual 2)
```
howsohazard: Do (* target_feature_residual target_feature_residual).  Also, why not just call LK_metric_exp?  Performance?

cademack: just was written in-line before. I don't think this is a tight enough loop to matter anyway. Should run F * (F-1) times max per analyze?

howsohazard (approved):

PR #577, opened 2025-07-22
24113: Improves validation logic for `goal_features_map` and `condition` parameters 

cademack:
Adds validation steps to ensure that all features referenced in goal features maps or condition maps are real/trained/system features in the Trainee. This new code will return errors when either of these parameters reference unknown features.

### DIFF
--- a/howso/input_validation.amlg
+++ b/howso/input_validation.amlg
@@ -118,7 +118,13 @@
 			parameter_name ""
 
 			;normally should not be overwritten
-			allowed_features (append !trainedFeatures !reactIntoFeaturesList [".session" ".session_training_index" ".series_index"])
+			allowed_features
+				(append
+					!trainedFeatures
+					!reactIntoFeaturesList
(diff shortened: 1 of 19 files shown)


jdbeel (commented):
Could this be something that's done automatically by `!ValidateParameters` if the function has condition or goal features as parameters? Obviously not required but could potentially make things cleaner (or the opposite 😆)

howso/input_validation.amlg
```diff
@@ -108,6 +108,73 @@
...
+	;verifies that the indices are either trained features (or built-in features)
+	;returns (null) if no error, string error if input is invalid
+	#!ValidateConditionMap
+	(declare
+		(assoc
+			given_map (assoc)
+			parameter_name ""
+
+			;normally should not be overwritten
+			allowed_features (append !trainedFeatures !reactIntoFeaturesList [".session" ".session_training_index" ".series_index"])
```
howsohazard: What about distance contributions and other anomaly detection features, case weights, time series, etc.?

howsohazard (approved):

PR #578, opened 2025-07-22
21409: Updates queries to use Amalgam's internal feature weight redistribution

cademack:


howso/residuals.amlg
```diff
@@ -1166,7 +1137,7 @@
 													(replace !queryDistanceTypeMap)
 													query_feature_attributes_map
 													(replace feature_deviations)
-													(null)
+													(replace features) ;weight selection
```
howsoRes: just `feature`. see removed lines 1077 and 1117 of previous code

howsoRes (approved):

PR #584, opened 2025-07-28
24166: Adds `sort_by` parameter to `#get_cases`

cademack:


howso/get_cases.amlg
```diff
@@ -263,6 +263,10 @@
 			;{ref "Precision"}
 			;string, default is 'exact', used only with 'condition' parameter, will find exact matches if 'exact' and similar cases if 'similar'.
 			precision "exact"
+			;{ref "SortBy"}
```
fulpm: Is this type name too generic? Maybe this should be something like `SortByFeature` instead.

howso/get_cases.amlg
```diff
@@ -336,13 +346,75 @@
...
 
+		(if (size sort_by)
+			(let
+				(assoc
+					sorting_features (map (lambda (get (current_value) "feature")) sort_by)
+				)
+
+				(assign (assoc
+					cases
+						(sort
```
howsoRes, line 363: might be worth adding a block of code to handle a single sort_by condition by itself using the query engine

howso/get_cases.amlg
cademack: I was just matching the name to the parameter - didn't think about it too hard. I could add Feature(s). I'd prefer to make it plural since you can sort by multiple features with it

fulpm: It should be singular since it represents one feature item in the list.

howsoRes (approved):

howso/get_cases.amlg
cademack: The schema it references is the list of maps, not just one

PR #598, opened 2025-08-13
24256: Adds `get_value_masses` endpoint to the Trainee API, MINOR

cademack:
Adds a new endpoint: `get_value_masses`. Which is designed to give users a way to get the unique values and their masses for each requested feature.

Example response payload:
```
{
        masses {
                        target {
                                        remaining 19
                                        values [
                                                        [0 29]
                                                        [1 27]
                                                ]
                                }
                }
}
```

### DIFF
--- a/howso/marginal.amlg
+++ b/howso/marginal.amlg
@@ -954,6 +954,8 @@
 
 
 	;outputs all unique values and the count of each value given the trained data and possible conditions
+	;if a weight feature is specified or auto-ablation is enabled, then the probability masses are returned rather
+	;than frequency of values.
 	#get_value_counts
 	(declare
 		; returns {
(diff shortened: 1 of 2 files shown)


howso/marginal.amlg
```diff
@@ -951,4 +951,136 @@
...
+						condition condition
+						precision precision
+					))
+				)
+		))
+
+		(declare (assoc
+			output_map
+				(map
+					(lambda
```
howsoRes: lambda and let on one line and un-indent everything in there once please

howso/marginal.amlg
```diff
@@ -951,4 +951,136 @@
...
+						(let
+							(assoc
```
howsoRes: single line assoc please:  `(assoc feature (current_index 1) )`

howsoRes (approved):

howso/marginal.amlg
```diff
@@ -951,4 +951,132 @@
...
+			condition (null)
+			;{ref "Precision"}
+			;default is 'exact', used only with 'condition' parameter, will find exact matches if 'exact' and similar cases if 'similar'.
+			precision "exact"
+			;{type "number" min 0}
+			;The minimum number of cases must hold a value for that value to be returned. Any values with frequency lesser than the
+			;specified minimum count have their frequencies added together and returned under the "remaining" key for the feature.
+			minimum_count_threshold 0
+			;{type "string"}
+			;name of case weight feature
```
howsohazard: What is the behavior if weight feature is used?  Does it sum up those instead?  Should be clear in the documentation.

PR #601, opened 2025-08-14
24257: Adds "return_context_values" flag to react and single_react, MINOR

cademack:
Adda a new flag to `#react`, `return_context_values`. When true, "context_values" and "context_features" will be returned in the react payload.

### DIFF
--- a/howso/react.amlg
+++ b/howso/react.amlg
@@ -513,6 +513,15 @@
 					)
 					(call !ComputePrecisionEpsilon)
 				)
+			output_context_features
+				(values (append
+					context_features
+					(if (= (null) desired_conviction)
+						(filter (lambda (not (contains_value action_features (current_value)))) preserve_feature_values)
+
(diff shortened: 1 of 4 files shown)


howso/react.amlg
```diff
@@ -510,6 +513,15 @@
 					)
 					(call !ComputePrecisionEpsilon)
 				)
+			output_context_features
+				(values (append
```
howsoRes: should this be in a (if return_context_values block, so that way output_context_features remains null if it's not needed?

howso/react.amlg
```diff
@@ -909,6 +932,15 @@
 					)
 					(call !ComputePrecisionEpsilon (assoc generated_uniques_list_map pre_generated_uniques_map))
 				)
+			output_context_features
+				(values (append
```
howsoRes: same comment as above

howsoRes (approved):

PR #610, opened 2025-08-25
24323: Fixes similarity and distance contributions computed with react_group alternated with react_into_features

howsoRes:
calling react_into_features stores computed DC into cases to be used in future similarity_conviction calls. however calling react_group a) may use a different feature set and b) will use filtering_queries ignoring other cases from each individual group, thus stored DC values cannot be used and both case DC and local data DC must be computed everytime

### DIFF
--- a/howso/conviction.amlg
+++ b/howso/conviction.amlg
@@ -273,6 +273,8 @@
 						(if (not (contains_value !reactIntoFeaturesList distance_contribution))
 							(accum_to_entities (assoc !reactIntoFeaturesList distance_contribution))
 						)
+
+						(assign_to_entities (assoc !storedDistanceContributionsFeatureSet (sort features) ))
 					)
 				)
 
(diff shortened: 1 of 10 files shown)


howso/conviction.amlg
```diff
@@ -273,6 +273,8 @@
 						(if (not (contains_value !reactIntoFeaturesList distance_contribution))
 							(accum_to_entities (assoc !reactIntoFeaturesList distance_contribution))
 						)
+
+						(assign_to_entities (assoc !storedDistanceContributionsFeatureSet (sort features) ))
```
howsohazard: How does this work if the user is requesting both case similarity conviction and group similarity conviction?  we should support both, among others

howsoRes: this is in react_into_features(), so this stores the feature set that was used to store DC into each case.  calling react_group w/ anything computes those requested stats fresh everytime since they are group-specific

howso/conviction.amlg
```diff
@@ -667,9 +677,15 @@
...
-					;not all cases have distance_contribution, must calculate distance_contribution for nearest neighbors
-					(!= (size (contained_entities (query_exists "distance_contribution") )) (call !GetNumTrainingCases))
+					(or
+						;react_group specified filtering_queries to filter out group case ids
+						(and react_group (size filtering_queries))
+						;or number of cases in datasets does not equal the number of cases with distance contributions,
+						;if not all cases have distance_contribution, must calculate distance_contribution for nearest neighbors
+						(!= (size (contained_entities (query_exists distance_contribution_feature) )) (call !GetNumTrainingCases))
+						;or features don't match feature set that was used to store DCs
+						(not (contains_index !storedDistanceContributionsFeatureMap (apply "concat" (sort features))) )
```
cademack, line 687: why not just check if distance_contribution_feature is contained in that map?

howsohazard (approved):

PR #623, opened 2025-09-03
24392: Improves performance of case residual conviction

howsoRes:


cademack (approved):

howso/details_residuals.amlg
```diff
@@ -221,20 +222,21 @@
...
+	;For full residuals flows outputs the assoc of feature -> predicted value
 	#!CalculateCaseResiduals
-	(declare
-		(assoc
-			case_residuals
+	(seq
+		(assign (assoc
+			case_residuals_map
 				(if robust_residuals
 					;for each feature, do 30 reacts with randomly selected contexts
```
howsohazard, line 234: Add comment stating that it is iterating over the keys of an assoc.  Why make an assoc, why not get unique values?

howso/details_residuals.amlg
```diff
@@ -327,43 +329,60 @@
...
 									)
-
-									;else this is a continuous feature, return the absolute difference between the expected and the actual
-									(abs (- expected_case_value (get reaction 0)))
-								)
+								]
 							)
 						))
-						action_features
+						(zip action_features)
```
howsohazard, line 361: Same as comment above

howso/details_residuals.amlg
howsohazard: Or, instead, make a comment stating what it's doing

howsohazard (approved):

PR #634, opened 2025-09-12
24448: Fixes issues in split-training flow where analysis/reduction could be triggered in the read-only op

cademack:


howso/scale.amlg
```diff
@@ -148,15 +148,22 @@
 		(declare (assoc
 			num_cases (call !GetNumTrainingCases)
 
+			;save the user-specified value for these flags to pass in the response for process-payload
+			user_skip_reduce_data skip_reduce_data
```
howsohazard: Don't use the word "user" here because we don't know it's a user.  Use "force" or name it after the particular flow or parameter.

howsohazard (approved):

PR #635, opened 2025-09-12
24434: Improves memory usage during analyze for some datasets with many features

howsoRes:


cademack (approved):

howso/mda_weight.amlg
```diff
@@ -341,128 +343,169 @@
...
-													(unzip (get feature_residuals_lists mda_feature_index) case_indices_without_feature)
-												))
-
-												;average with residuals
-												(generalized_mean (map
-													(lambda (last (current_value)))
-													(unzip (get feature_residuals_lists mda_feature_index) case_indices_with_feature)
-												))
-											)
+			(accum (assoc
```
howsohazard: Rather than accum'ing start and end, couldn't you move the logic up to (range start end) and just do some arithmetic there based on iteration numbel

howsohazard (approved):

PR #639, opened 2025-09-18
24481: Adds logic to react_series to enforce minimum nonzero time-deltas when using a datetime time feature

cademack:


howso/attribute_maps.amlg
```diff
@@ -17,6 +17,28 @@
...
 	)
 
+	;A list of tuples containing a datetime format symbol and its granularity in seconds
+	; sorted in increasing granularity to help determine the granularity of a datetime format.
+	#!DatetimeSymbolPrecisions
+	[
+		["%S" 1] ;Second padded
+		["%M" 60] ;Minute padded
+		["%H" (* 60 60)] ;Hour (12)
+		["%H" (* 60 60)] ;Hour (24)
```
howsoRes, line 32: duplicate, just make comment that it's 12/24 .. or make it lower case h?

howsoRes (approved):

howso/attribute_maps.amlg
```diff
@@ -42,6 +64,15 @@
 								;set default of en_US if locale is not specified
 								"locale" (replace locale)
 								"default_time_zone" (if default_time_zone default_time_zone "UTC")
+								"min_nonzero_delta"
```
howsohazard, line 86: How does this handle epoch data with fractional seconds?

howsohazard (approved):

PR #647, opened 2025-10-03
24515: Refactors format of derived and custom code to not use labels, MAJOR

howsoRes:
replaces instances of feature references in derivation code from the generic format of **#"feature name" <lag_value>**
to: **(call value{ feature "feature_name" _lag lag_value_ })** where lag and lag_value are optional and only needed if lag > 0
e,g.:
 "(+ #\\"my feature\\" 0  #feat_b 2)"   
 is now: 
 "(+ (call value {feature \\"my feature\\"}) (call value {feature \\"feat_b\\" lag 2}) )"

### DIFF
--- a/howso/contributions.amlg
+++ b/howso/contributions.amlg
@@ -734,7 +734,7 @@
 					[(call_sandboxed
 						(call !ParseDerivedFeatureCode (assoc
 							code_string (get !featureAttributes [action_feature "derived_feature_code"])
-							get_val (lambda (if (= lag 0) (get case f) ) ) ;;;^^^TODO: (lambda lag) and (lambda f) ?
+							get_val (lambda (if (= lag 0) (get case feature) ) )
 						))
 						{"case" (append case_values_map values_for_derivation_map)}
 						!sandboxedComputeLimit !sandboxedMemoryLimit !sandboxedOpcodeDepthLimit (false)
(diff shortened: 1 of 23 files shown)


howso/contributions.amlg
```diff
@@ -739,15 +734,7 @@
...
-							label_to_code
-								(lambda
-									(if (= (lambda label_value) 0)
-										;pull the feature value
-										(get case (lambda label_name))
-
-										(null)
-									)
-								)
+							!get_val (lambda (if (= lag 0) (get case f) ) ) ;;;^^^TODO: (lambda lag) and (lambda f) ?
```
cademack: Is there something else needed here?

howso/custom_codes.amlg
```diff
@@ -291,53 +289,42 @@
 		)
 	)
 
-	;constrains code to the specified list of allowed_opcodes
-	; then it finds any node that has a label and replaces it by a copy of label_to_code,
-	; replacing label_value with the value at the label and label_name with the label name
+	;rewrites the code in code_string by replacing all instances of (call get_val { f <feature> lag <lag> }) with inlined
```
howsohazard: should be get_value and feature instead of f  (replace everywhere)

howso/contributions.amlg
howsoRes: no forgot to remove todo

cademack (approved):

PR #649, opened 2025-10-08
24487: Adds clustering option to react_into_features, MINOR

howsoRes:
populates a new feature called ".cluster_id" to cases

### DIFF
--- a/howso/clustering.amlg
+++ b/howso/clustering.amlg
@@ -101,7 +101,7 @@
 						(apply "max" (unzip
 							case_to_dc_map
 							;cases in this cluster
-							(filter (lambda (= cluster_id (current_value))) case_cluster_map)
+							(indices (filter (lambda (= cluster_id (current_value))) case_cluster_map))
 						))
 					))
 					;unique cluster ids
@@ -196,6 +196,8 @@
 			)
 			case_to_sc_map
 		)
+
+		(print  case_cluster_map " " (size (values case_cluster_map (true)))" "  (size case_to_sc_map))
 	)
 
 
--- a/howso/clustering.amlg
+++ b/howso/clustering.amlg
@@ -65,7 +65,9 @@
 				)
 			case_id (null)
 			cluster_id 1
-			border_cases []
+
+			;cases in one cluster and close to another but aren't reciprocated by those in the other
+			bridging_cases []
 		))
 
 		;grab the next unclustered case, cluster it and its valid neighbors
@@ -268,9 +270,11 @@
 							(assign (assoc cluster_map (remove cluster_map eid) ))
 						)
 
-						(if (not (contains_value border_cases case_id) )
-							(accum (assoc border_cases case_id ))
-						)
+						;case is in one cluster and close to another but isn't reciprocated by those in the other
+						;what should be done with bridging_cases ?
+						; (if (not (contains_value bridging_cases case_id) )
+						; 	(accum (assoc bridging_cases case_id ))
+						; )
 					)
 				)
 			)
--- a/howso/clustering.amlg
+++ b/howso/clustering.amlg
@@ -132,28 +132,24 @@
 					;all neighbors are unclustered, do another pass since those neighbors may be assigned to this cluster during this pass
 					(if (= 0 (size neighbor_cluster_map))
 						(assign (assoc another_pass (true) ))
-					)
 
-					(declare (assoc
-						cluster_id
-							(if (and
-									(size neighbor_cluster_map)
-									(= 1 (size (values neighbor_cluster_map (true))))
-								)
-								(first (values neighbor_cluster_map))
-							)
-
-					))
-					(if cluster_id
-						;if the DC of this case isn't too far away, i.e., is less than a multiple of the largest DC of this cluster
-						;add this case that cluster as well
+						;else there's exactly one cluster id among the neighbors
+						(= 1 (size (values neighbor_cluster_map (true))))
+						;if the DC of this case isn't too far away, i.e., is less than clustering_inclusion_relative_threshold of the
+						;largest DC of this cluster, add this case that cluster as well
 						(if
 							(<
 								(get case_to_dc_map (current_value))
-								(* 1.5 (get cluster_max_dc_map cluster_id))
+								(*
+									clustering_inclusion_relative_threshold
+									(get cluster_max_dc_map (first (values neighbor_cluster_map)))
+								)
 							)
 							(let
-								(assoc case_id (current_value 1) )
+								(assoc
+									case_id (current_value 1)
+									cluster_id (first (values neighbor_cluster_map))
+								)
 								(accum (assoc case_cluster_map (associate case_id cluster_id) ))
 								(assign (assoc
 								  	case_to_sc_map (remove case_to_sc_map case_id)
@@ -200,7 +196,6 @@
 			)
 			case_to_sc_map
 		)
-
 	)
 
 
--- a/howso/clustering.amlg
+++ b/howso/clustering.amlg
@@ -36,6 +36,7 @@
 			next_case_index 0
 		))
 
+		;cache distances to all neighbors
 		(declare (assoc
 			all_cases_neighbors_map
 				||(map
@@ -66,8 +67,8 @@
 			cluster_id 1
 			border_cases []
 		))
-		(print "Cached distances to neighbors.\n")
 
+		;grab the next unclustered case, cluster it and its valid neighbors
 		(while (!= (null) next_case_index)
 			(assign (assoc
 				case_id (call !GetNextCaseToCluster)
@@ -202,7 +203,6 @@
 			case_to_sc_map
 		)
 
-		(print cluster_map "\n" "num_unclustered " (size case_to_sc_map) "\n" )
 	)
 
 
@@ -212,7 +212,7 @@
 		(if (contains_index case_to_sc_map (get case_ids_by_sc (+ (current_index) next_case_index)))
 			(let
 				(assoc case_id (get case_ids_by_sc (+ (current_index 1) next_case_index)) )
-				;;;^^^ tweak
+				;;;^^^ tweak?
 				(if ;(or
 						(< (get case_to_sc_map case_id) 1)
 					;	(> (get case_to_dc_map case_id) !averageCaseDistanceContribution)
@@ -288,7 +288,7 @@
 					neighbors
 						(filter
 							(lambda
-								;;;^^^ tweak
+								;;;^^^ tweak?
 								;(or
 									(>= (get case_to_sc_map (current_value)) expansion_sc_threshold)
 								;	(<= (get case_to_dc_map (current_value)) (* (/ 1 expansion_sc_threshold) !averageCaseDistanceContribution))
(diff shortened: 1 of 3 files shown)


howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+#!GetNextCaseToCluster
+	(while (< (+ (current_index) next_case_index) num_cases)
+		(if (contains_index case_to_sc_map (get case_ids_by_sc (+ (current_index) next_case_index)))
+			(let
+				(assoc case_id (get case_ids_by_sc (+ (current_index 1) next_case_index)) )
+
+				;stop pulling the next case if they have low SC
+				(if ;(or
+						(< (get case_to_sc_map case_id) 1)
+					;	(> (get case_to_dc_map case_id) !averageCaseDistanceContribution)
```
cademack: Are we removing this second condition?

howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+			)
+			case_to_sc_map
+		)
+
+	)
+
+
+;output case_id of next case, remove case_id from case_to_sc_map
+#!GetNextCaseToCluster
+	(while (< (+ (current_index) next_case_index) num_cases)
```
cademack, line 205: I don't think our pattern is to indent the first line after the label (applies to the whole module I think)

howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+				(assign (assoc
+					case_to_sc_map (remove case_to_sc_map case_id)
+					next_case_index (+ 1 (current_index 1) next_case_index)
+				))
+				(conclude case_id)
+			)
+		)
+	)
+
+#!RecursivelyLabelCluster
```
cademack: Needs description (mentioning the parameters passed in should probably be here too)

howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+										case_cluster_map
+									)
+							))
+						)
+
+						;case is in one cluster and close to another but isn't reciprocated by those in the other
+						;what should be done with bridging_cases ?
+						; (if (not (contains_value bridging_cases case_id) )
+						; 	(accum (assoc bridging_cases case_id ))
+						; )
```
cademack, line 272: Leaving a TODO here?

howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+		(map
+			(lambda
+				(if (and
+						(!= (null) (get case_cluster_map (current_value)))
+						(!= cluster_id (get case_cluster_map (current_value)))
+					)
+					;this case is reciprocated as a neighbor of this other-cluster case
+					(if (contains_index (get all_cases_neighbors_map (current_value)) case_id)
+						(let
+							(assoc eid (get case_cluster_map (current_value 1))  )
```
cademack: Eid is a holiday, not sure what it means here. This is the cluster ID of the a neighbor case who is also a neighbor of the original case? 
I'd like a more informative variable name if possible

howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+
+							(assign (assoc
```
cademack, line 259: I think a comment of the implication of this assign would be informative

howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+				;leave only neighbors with "good" SC and DC values
+				(assign (assoc
+					neighbors
+						(filter
+							(lambda
+								;;;^^^ tweak?
+								;(or
+									(>= (get case_to_sc_map (current_value)) expansion_sc_threshold)
+								;	(<= (get case_to_dc_map (current_value)) (* (/ 1 expansion_sc_threshold) !averageCaseDistanceContribution))
+								;)
```
cademack: Empty comment to demand something be done here before merge

howso/clustering.amlg
```diff
@@ -0,0 +1,388 @@
...
+								)
+								(declare (assoc unique_neighbor_clusters (values other_neighbor_clusters_map (true)) ))
+
+								;this neighbor's neighbors are in a different cluster
+								;interpolate based on influence weight to see which cluster this neighbor should be assigned to
+								(if (or
+										(> (size unique_neighbor_clusters) 1)
+										(and
+											(size unique_neighbor_clusters)
+											(!= cluster_id (first unique_neighbor_clusters))
```
cademack, line 322: Is it impossible for cluster_id to be in the list of unique_neighbor_clusters?

cademack: err, maybe thats not describing my concern well. I don't get why checking the (first) here makes sense

howso/conviction.amlg
```diff
@@ -38,6 +38,13 @@
 			;true or string, if true will use default value of "influence_weight_entropy"
 			influence_weight_entropy (null)
 			;{type "boolean"}
+			;if true will cluster and store cluster ids into ".cluster_id".
+			;will compute and overwrite distance contributions and similarity convictions
```
cademack: Does this overwrite behavior interact with the overwrite flag? If not, it probably should

howso/conviction.amlg
```diff
@@ -38,6 +38,13 @@
...
+			clustering (null)
+			;{type "number" exclusive_min 0}
+			;similarity conviction threshold of cases considered for expansion of cluster
```
cademack: This should be a bit more descriptive IMO

howso/conviction.amlg
```diff
@@ -38,6 +38,13 @@
 			;true or string, if true will use default value of "influence_weight_entropy"
 			influence_weight_entropy (null)
 			;{type "boolean"}
+			;if true will cluster and store cluster ids into ".cluster_id".
+			;will compute and overwrite distance contributions and similarity convictions
+			clustering (null)
```
cademack, line 43: If we want this to match the RiF pattern, this parameter should be something like "cluster_id" and have that default as you have described, but also optionally accept a string name for the user to store the ID values under.

howso/clustering.amlg
howsoRes: it is very possible, the or statements checks, either A) there are more than one: in which case the neighbor may be assigned to that other one or B) there's only one and it doesn't match this cluster_id

howso/conviction.amlg
howsohazard: Agreed this needs more thought.  The generic parameter "feautres" above lists itself as only being for calculating convictions in the comment, but even that is not clear how I would use it for multiple types of conviction.

howso/clustering.amlg
```diff
@@ -0,0 +1,415 @@
...
+							)
+
+					))
+					(if cluster_id
+						;if the DC of this case isn't too far away, i.e., is less than a multiple of the largest DC of this cluster
+						;add this case that cluster as well
+						(if
+							(<
+								(get case_to_dc_map (current_value))
+								(* 1.5 (get cluster_max_dc_map cluster_id))
```
howsohazard: Why 1.5?  What is this magic number and why is it good?

howso/clustering.amlg
```diff
@@ -0,0 +1,415 @@
...
+						cluster_id
+							(if (and
+									(size neighbor_cluster_map)
+									(= 1 (size (values neighbor_cluster_map (true))))
+								)
+								(first (values neighbor_cluster_map))
+							)
+
+					))
+					(if cluster_id
```
howsohazard: Can this be combined into one if / else switch statement so that cluster_id doesn't need to be defined at all?  Should be more efficient and probably slightly less code.

howso/clustering.amlg
```diff
@@ -0,0 +1,415 @@
...
+						(let
+							(assoc other_cluster_id (get case_cluster_map (current_value 1))  )
+
+							;for all cases that have other_cluster_id, overwrite it with this cluster_id
+							(assign (assoc
+								case_cluster_map
+									(map
+										(lambda (if (= other_cluster_id (current_value)) cluster_id (current_value)) )
+										case_cluster_map
+									)
```
howsohazard, line 264: Commented out code -- should remove or add todo

howso/clustering.amlg
howsoRes: It's somewhat arbitrary, but acquired via good results with testing.  Somewhat like the magic numbers in garbage collection in Amalgam.   We could parameterize this as well if it's a concern. Let me know.

howso/clustering.amlg
```diff
@@ -0,0 +1,415 @@
...
+		;leave only unclustered neighbors
+		(assign (assoc
+			neighbors (filter (lambda (= (null) (get case_cluster_map (current_value)))) neighbors)
+		))
+		(if (size neighbors)
+			(seq
+				;leave only neighbors with "good" SC and DC values
+				(assign (assoc
+					neighbors
+						(filter
```
howsohazard, line 288: Commented out code

howso/clustering.amlg
howsoRes: updated comment for 'features' parameter

howsohazard: Sure, but those magic numbers in Amalgam are substantiated by many decades of research and common in GC systems :-).  Plus there are comments in the functions that use them.  Don't tell me here, describe it in the code.

howso/conviction.amlg
howsoRes: updated comment for 'features' parameter

howso/conviction.amlg
```diff
@@ -9,7 +9,7 @@
 	(declare
 		(assoc
 			;{type "list" values "string"}
-			;list of features for which to calculate conviction, will default to trainee's default features if unspecified
+			;list of features used to do all the computations, will default to trainee's default features if unspecified
```
howsohazard, line 12: How does it zip the order of these values to the type of computation?  I can't tell from this description.

howsohazard: Also, it says "all the computations" but clustering doesn't use this.

howsoRes: it does, it uses features for clustering as well - dc, sc and distances are all computed w the specified features

howso/conviction.amlg
```diff
@@ -38,6 +38,14 @@
 			;true or string, if true will use default value of "influence_weight_entropy"
 			influence_weight_entropy (null)
 			;{type "boolean"}
+			;if true will cluster and store cluster ids into ".cluster_id".
+			;will compute and store distance contributions and similarity convictions while respecting the overwrite flag.
+			clustering (null)
+			;{type "number" exclusive_min 0}
+			;similarity conviction threshold of cases considered for expansion of a cluster, only cases with similarity conviction
+			;equal to or greater than this value will be considered to be clustered in the same cluster as their neighbors.
+			clustering_expansion_threshold 0.5
```
howsohazard, line 47: This is for clustering, so name of parameter should reflect that.

howsoRes: outdated comment?

howsohazard: Ah yeah, github in VSCode said it refreshed but it's still stuck with some older version :-/.

cademack (approved):

howso/conviction.amlg
```diff
@@ -38,6 +38,18 @@
...
+			;{type "number" exclusive_min 0}
+			;The initially unclustered candidate cases' distance contirbution needs to be less than this value times the
```
jackx111: contirbution -> contribution

PR #650, opened 2025-10-08
24538; Adds support for derived stationary features in time series, MINOR

howsoRes:
specifying a feature with an 'auto_derive_on_train' with derive_type='custom' and no time_series block marks it a stationary feature.


howsohazard (approved):

cademack:
Not a fan of adding a "stationary_features" field to the TS features map that really is just a list of these special derived stationary features. I feel the list should be renamed or it should include all features without a time-series block defined

PR #654, opened 2025-10-13
24569: Adds exclusive condition for nominal features, MINOR

howsoRes:


howso/get_cases.amlg
```diff
@@ -691,9 +692,21 @@
 							(contains_value (list "number" "string") (get_type_string cond_value))
 							(query_equals feature cond_value)
 
-							;else this is a list of values, for nominals use query_among, for continuous use query_between
+							;else this is a list (or assoc for nominals) of values
 							(if (or (contains_index !nominalsMap feature) (contains_index !ordinalStringToOrdinalMap feature))
-								(query_among feature cond_value)
+								;an assoc with the key 'exclude' blacklists the specified value(s)
+								(if (= "assoc" (get_type_string cond_value))
```
cademack: why use (get_type_string)? is this better than (~)?

cademack (approved):

howso/get_cases.amlg
howsoRes: no difference, just going with what was already in the method, i'll update

PR #656, opened 2025-10-14
24579: Fixes incorrect local full residuals for datasets with many nulls

howsoRes:


howso/details_residuals.amlg
```diff
@@ -27,6 +27,76 @@
...
+
+								;if there are less than 2 non-null values, need to find more local cases with non-nulls
+								(if (< (size local_not_null_cases) 2)
+									(append
+										local_not_null_cases
+										(contained_entities
+											filtering_queries
+											(query_not_equals feature (null))
+											(query_nearest_generalized_distance
+												(- 2 (size local_not_null_cases))
```
howsohazard: Expand to not include existing case AND make sure the minimum is min_k.  Apply to other spot as well, and comment, on minimum.

howso/residuals.amlg
```diff
@@ -1075,7 +1081,11 @@
 											)
 									))
 								))
-								case_ids
+								(if (size local_feature_case_id_map)
```
cademack, line 1085: (contains_index) might be slightly more appopriate/safe here than (size)

howsoRes: not really - if it's defined, it contains all features, otherwise it won't be defined

howsohazard (approved):

PR #667, opened 2025-10-22
24619: Adds support for a new time-series type: "covariate", MINOR

cademack:
This PR adds support for a new time-series type, "covariate".

This joins the list of "rate" and "delta" as valid time-series types for continuous time-series features. Rather than use a rate/delta and the lag to determine the next feature value in a series, feature specified as "covariate" have their next values in the series predicted directly via interpolation (still using as many lags as specified as contexts). 

These changes also make an adjustment to the dependent features logic that automatically pulls deltas/rates into the dependent relationships of their parents.


howsohazard (approved):

howso.amlg
```diff
@@ -611,10 +611,11 @@
 					"series_id_features" (list)
 					;list of all features that  need to be derived in the order they should be derived
 					"ts_derived_features" (list)
-					;flag, set to true if series has values that explicitly denote the end of a series
-					"series_has_terminators" (false)
-					;flag, set to true if a series must end on a terminator value
-					"stop_on_terminator" (false)
+					;list of feature names of features with stationary/backfill logic supported
```
howsohazard, line 614: Needs a unit test

howsohazard (approved):

howsoRes (approved):

PR #674, opened 2025-10-28
24649: Improves logic for series-termination and initial residuals

cademack:


howso/residuals.amlg
```diff
@@ -140,7 +140,13 @@
 							;else set nominal lower residual bound to:  gap / (num_cases + 1)
 							;multiply (current_value) by 2 because (current_value) is the 'half gap' value
 							(contains_index !nominalsMap (current_index))
-							(/ (* 2 (current_value)) (+ 1 num_training_cases))
+							(/
```
howsohazard, line 144: Add to the comment that nulls are ignored so only actual values are counted

howsoRes (approved):

PR #677, opened 2025-11-04
24467: Adds `index_offset` parameter to train, MINOR

cademack:


cademack:
Still need to verify this works through TS ablation... do not merge yet

PR #696, opened 2025-11-21
24759: Refactors distribution of reduced cases to use multithreading

howsoRes:


howso/ablation.amlg
```diff
@@ -87,7 +84,12 @@
 		))
 
 		;output tuple of: [ entropy, most influential case id, influence of most influential, influence of least influential, is_duplicate]
-		[ influence_weight_entropy (first (first local_cases_pairs)) closest_influence (last (last local_cases_pairs)) is_duplicate ]
+		[
+			(entropy (normalize (last local_cases_pairs)))
+			(first (first local_cases_pairs))
+			closest_influence
+			(last (last local_cases_pairs)) is_duplicate
```
cademack: is there a reason that `is_duplicate` doesn't get its own newline?

howso/update_cases.amlg
```diff
@@ -850,146 +850,225 @@
...
+							(map (lambda (last (current_value))) (values distributed_cases_maps))
+						)
+				)
+
+				;distribute the corresponding portion of this case's weight based on the neighbor's influence
+				(map
+					(lambda
+						(accum_to_entities (current_index) (associate distribute_weight_feature (current_value 1)) )
+					)
+					accumulated_individual_weights_map
```
cademack: I wouldn't hate if we just inline the reduce in the map rather than use a let/assoc

cademack (approved):

howso/update_cases.amlg
```diff
@@ -1063,6 +1003,60 @@
...
+		(lambda
+			(apply "+"
+				(filter (map
+					(lambda (get (current_value) [(current_index 2) index]))
+					list_of_assocs
+				))
+			)
+		)
+		;assoc of only the unique ids among all the lists
+		(zip (indices (apply "append" list_of_assocs)))
```
cademack: if we do `(zip (indices _))` here, should we not just use `_` in its place? The zip/indices don't seem needed since we will still iterate over each index without those steps

howso/update_cases.amlg
```diff
@@ -1063,6 +1003,60 @@
...
+		(lambda
+			(apply "+"
+				(filter (map
+					(lambda (get (current_value) (current_index 1)))
+					list_of_assocs
+				))
+			)
+		)
+		;assoc of only the unique ids among all the lists
+		(zip (indices (apply "append" list_of_assocs)))
```
cademack: my other comment also applies here

howso/update_cases.amlg
```diff
@@ -956,75 +956,26 @@
...
 				)
+				;reduce all the closest cases maps into one map of individual case -> total accumulated weight
+				(call !ReduceAssocsListValuePair (assoc
+					list_of_assocs (map (lambda (last (current_value))) (values distributed_cases_maps))
+					;weight is stored as the first in a pair
+					index 0
+				))
+				;reduce all the closest cases maps into one map of individual case -> total accumulated masses
+				(call !ReduceAssocsListValuePair (assoc
+					list_of_assocs (map (lambda (last (current_value))) (values distributed_cases_maps))
```
cademack, line 975: We don't need (last) of each cur value if using ReduceAssocsListValuePair right?

cademack: Oh I think I see, never mind

PR #700, opened 2025-12-02
24776: Adds value_robust_prediction_contributions to react_aggregate, MAJOR

howsoRes:
react_aggregate() changes:
  &nbsp;  added 'value_robust_prediction_contributions' details
renamed input keys: 
  &nbsp;  value_robust_ac_features -> value_robust_contributions_features 
  &nbsp;   value_robust_ac_action_features -> value_robust_contributions_action_features 
  &nbsp;  value_robust_ac_max_num_buckets   -> value_robust_contributions_num_buckets 
renamed details output key to: 'value_robust_contributions'

### DIFF
--- a/howso/react_aggregate.amlg
+++ b/howso/react_aggregate.amlg
@@ -93,7 +93,7 @@
 			;{type "number" min 0}
 			;Total sample size of dataset to use (using sampling with replacement) for feature_robust_accuracy_contributions,
 			;estimated_residual_lower_bound and value_robust_accuracy_contributions.
-			;	Defaults to the smaller of 150000 or (6321 * num_context_features), for value_robust_accuracy_contributions defaults to 100000.
+			;	Defaults to the smaller of 150000 or (6321 * num_context_features), for value robust contribution flows defaults to 100000.
 			num_robust_accuracy_contributions_samples (null)
 			;{type "number" min 0}
 			;Percent threshold used to dynamically limit the number of samples used to determine robust accuracy contributions.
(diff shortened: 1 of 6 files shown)


howso/value_contributions.amlg
```diff
@@ -451,15 +539,40 @@
...
 				)
-			"ac_values" (unzip corresponding_acs sorted_indices)
-		}
-	)
+		))
 
+		;split the lists of all values into corresponding lists for output:
+		;just the unique feature values and their corresponding AC and/or PC values
+		(if (and compute_ac compute_pc)
+			{
```
cademack, line 550: I feel the `(append {..} (if blah {..} {}) ...)` pattern here may be a bit more succinct/less repetitive

howso/react_aggregate.amlg
```diff
@@ -117,14 +117,15 @@
 			;	be accumulated into a single value of all insignificant predictions for the class and removed from the confusion matrix.
 			confusion_matrix_min_count 15
 			;{type "list" values "string"}
-			;features for which to determine classes or values that cause the highest accuracy contribution to the specified 'value_robust_ac_action_feature'
-			value_robust_ac_features (null)
+			;features for which to determine classes or values that cause the highest accuracy and/or prediction contribution to the
+			;specified 'value_robust_contribution_action_features'
+			value_robust_contribution_features (null)
```
cademack: I think all the parameter names should have `contribution*s*`. Seems more consistent with the existing parameter names to me at least

howso/value_contributions.amlg
howsoRes: i did this specifically because an append would be less performant wrt to memory usage due to copying a possible very large assoc of all value combinations

howso/react_aggregate.amlg
```diff
@@ -199,7 +200,11 @@
 			;		context_num_samples: optional, limit on the number of context cases when 'context_condition_precision' is set to 'similar'.
 			;			If null, will be set to k. default is null
 			;		value_robust_accuracy_contributions: optional, none/true/false.
-			;			Perform a focused computation to determine how all the individual values of specified 'value_robust_ac_features' affect the accuracy of 'value_robust_ac_action_feature'.
+			;			Perform a focused computation to determine how all the individual values of specified 'value_robust_contribution_features' affect the accuracy of 'value_robust_contribution_action_features'.
+			;			Outputs under 'value_robust_contributions'
+			;		value_robust_prediction_contributions: optional, none/true/false.
```
howsohazard, line 205: should also have value_robust_accuracy_contributions separately to be consistent with the rest of the API

howsoRes: discussed this with Matt, he agreed that having this general one key is not inconsistent with the rest of the API where we have keys in output that don't exactly match the input. Also this reduces the possible payloads and the need to duplicate large outputs (all possible feature values, especially if there are many combinations)

howso/react_aggregate.amlg
howsoRes: done

howso/react_aggregate.amlg
fulpm: @howsohazard `value_robust_accuracy_contributions` looks to be a pre-existing parameter that has not changed in this PR, so this does look consistent.

fulpm (approved):

PR #715, opened 2025-12-15
24849: Adds `tokenizable_string` as an allowed `original_type` value in the feature attributes, MINOR

apbassett:


mkoistinen (approved):
LGTM

PR #723, opened 2025-12-19
24821: Refactors reduce_data() to reduce via normalized references, MAJOR

howsoRes:
removes 'influence_weight_entropy_threshold ' and adds 'reduce_max_cases' instead  to reduce_data()


howsoRes:
superceeded by 24821-reduce-refactor

PR #726, opened 2025-12-30
24891: Adds recursive_matching as feature attribute for code features, MINOR

howsoRes:


howso/attributes.amlg
```diff
@@ -49,6 +49,12 @@
 			;				'string_mixable', 'formatted_time' and 'formatted_date_time' are valid only when type is continuous
 			;				'boolean' is valid only when type is nominal
 			;
+			;	'recursive_matching': boolean, only applicable to code features (when 'data_type' is one of json/yaml/amalgam).
+			;				If unspecified, defaults to false for 'json' and 'yaml' features and true for 'amalgam' features.
+			;				When specified, this flag is passed down to any code-related operations (e.g., mixing or computing edit distances)
```
howsohazard: "passed down"?  Remove that phrase, doesn't make sense.
"when false,..." -> "when false, will operate on positional matches without considering recursion, which will yield better and faster results if the schema of the semistructured data is not recursive."

howso/types.amlg
```diff
@@ -75,6 +75,18 @@
...
 							)
 						data_type {ref "FeatureDataType"}
+						recursive_matching
+							(assoc
+								type "boolean"
+								description
+									(concat
+										"Only applicable to code features (when 'data_type' is one of json/yaml/amalgam).\n"
+										"If unspecified, defaults to false for 'json' and 'yaml' features and true for 'amalgam' features.\n"
+										"When specified, this flag is passed down to any code-related operations (e.g., mixing or computing edit distances)\n"
```
howsohazard: Update w/ text from above

cademack (approved):

howsohazard (approved):

PR #731, opened 2026-01-07
24917: Improves handling of time-series derived features in residual convictions flows

cademack:


howso/details_residuals.amlg
```diff
@@ -259,6 +259,28 @@
 			inactive_features_zeros_map (assoc)
 		))
 
+		(if (and
```
howsoRes, line 262: can you refactor this whole thing be in the `(declare (assoc undefined_ts_features` and change the seq inside the if statement to a let, store ts_feature_lag_amount_map into the locally scoped let-assoc, and output of the let should be the filter statement.

howsoRes (approved):

PR #738, opened 2026-01-14
24940: Updates reduce_data to remove entire series if not enough series cases remaining

howsoRes:
In reduce flow, if there are <= 3 cases for a series left when attempting to remove a start/end case, the entire series should be removed

### DIFF
--- a/howso/ablation.amlg
+++ b/howso/ablation.amlg
@@ -775,7 +775,12 @@
 						(accum (assoc
 							cases
 								(apply "append" (map
-									(lambda (contained_entities (current_value)))
+									(lambda
+										(contained_entities
+											(query_not_in_entity_list cases)
+											(current_value)
+										)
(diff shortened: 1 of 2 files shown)


howso/ablation.amlg
```diff
@@ -725,15 +725,62 @@
...
+								(query_not_equals ".series_index" 0)
+							)
+					))
+
+					;there were series that will need to be entirely removed, add all those series cases for removal
+					(if (size entire_series_removal_id_queries)
+						(accum (assoc
+							cases
+								(apply "append" (map
+									(lambda (contained_entities (current_value)))
```
cademack: should we append the (query_not_in_entity_list cases) here? Otherwise it seems to me that `cases` *could* have duplicate case ids otherwise

cademack (approved):

PR #749, opened 2026-01-26
24956: Updates derivation logic of #train to not incorrectly rederive values of TS data that has undergone reduction

cademack:


howso/derive_features.amlg
```diff
@@ -52,40 +52,53 @@
 					derived_features
 				)
 			time_feature_delta (concat "." !tsTimeFeature "_delta_1")
-		))
 
-		;derive all the lag features in one pass
-		(call !DeriveLagFeatures (assoc
-			features features
-			lag_features lag_features
+			has_reduced (size (contained_entities (query_exists !autoAblationWeightFeature) (query_not_equals !autoAblationWeightFeature 1.0)))
```
howsoRes: you don't need the query_exists, because the query_not_equals will implicitly only consider those cases that have that feature to do the value comparison.

howso/derive_features.amlg
```diff
@@ -493,6 +549,17 @@
 						)
 				))
 
+				(if is_missing_cases
+					(declare (assoc
+						case_id_to_idx_map
+							(zip
+								series_case_ids
+								(range 0 (- (size series_case_ids) 1))
```
howsoRes: replace this range call  with `(indices series_case_ids)`

howso/derive_features.amlg
```diff
@@ -464,6 +478,29 @@
...
+											(query_equals ".reverse_series_index" 0)
+										))
+										".series_index"
+									)
+								)
+						))
+
+
+						(declare (assoc
+							is_missing_cases (!= (size series_case_ids) num_cases_trained)
```
howsoRes: this check will fail if retraining the exact amount that was removed for this series.  a better check is something like `(size (contained_entities (query_in_entity_list series_case_ids) (query_not_exists ".series_index")) ` that checks whether there are cases in the series that haven't been derived yet

howso/derive_features.amlg
```diff
@@ -493,6 +549,17 @@
...
 				))
 
+				(if is_missing_cases
+					(declare (assoc
+						case_id_to_idx_map
+							(zip
+								series_case_ids
+								(range 0 (- (size series_case_ids) 1))
+							)
+						new_series_case_ids (indices (keep (zip series_case_ids) new_case_ids))
```
howsoRes: probably want to put this in its own (declare) block and replace (zip series_case_ids) with 'case_id_to_idx_map' that you just declared above, in case each series is very long, it won't have to re-zip all these case ids again when all you need is the indices.

howso/derive_features.amlg
howsoRes: make sure to update all the instances of is_missing_cases

cademack: Yeah this former check definitely had its issues - I've updated it with something that should be more robust/correct

howso/custom_codes.amlg
```diff
@@ -80,6 +81,54 @@
...
+						))
+
+
+						(declare (assoc
+							is_missing_cases
+								;if the number of old series cases is less
+								;than indicated by the largest series index of an old series case
+								(and
+									;this value will be null if there are no previous cases trained before this derivation
+									num_cases_prev_trained
```
howsoRes, line 119: num_cases_prev_tained will always be true because the logic for it is (+ 1 ... ), so this check isn't valid

cademack: the check is to catch is if its null - which is really when there are no series case ids that are not also new case ids, so the inner contained entities call returns an empty list.

howsoRes (approved):

PR #751, opened 2026-01-28
24986: Adds performance improvements to value oriented details in react_aggregate, MINOR

cademack:


howso/value_contributions.amlg
```diff
@@ -96,6 +96,183 @@
...
+		))
+
+		;create an assoc of feature -> all unique values (classes or buckets)
+		(declare (assoc
+			feature_value_lists_map
+				(zip
+					value_robust_contributions_features
+					(map
+						(lambda (let
+							(assoc f_idx (current_index 1))
```
howsoRes: this let and assoc seem unnecessary, you can just have the if statement in the lambda

howso/value_contributions.amlg
```diff
@@ -96,6 +96,183 @@
...
+		))
+
+		(declare (assoc
+			num_required_cases (max (min 30 (* (call !GetNumTrainingCases) 0.05)) 10)
+		))
+
+		;define if marginal values are significant enough to continue on with .true if value is significant, .false otherwise
+		(declare (assoc
+			marginal_feature_value_skip_map
+				(map
```
howsoRes: add a parallel bars || here?  or on the inner map ?

howso/value_contributions.amlg
```diff
@@ -96,6 +96,183 @@
...
+										;all unique classes
+										(size (contained_entities
+											(query_in_entity_list case_ids)
+											(query_equals (current_index 1) (current_index))
+										))
+
+										;all unique buckets
+										(size (contained_entities
+											(query_in_entity_list case_ids)
+											(query_between
```
howsoRes: for easier debugging, please comment what each of these curent_index with their offsets refer to here, it's getting a little confusing

howso/value_contributions.amlg
```diff
@@ -96,6 +96,183 @@
...
+					feature_value_lists_map
+				)
+		))
+
+		;filter case ids based on marginal value skip map
+		(assign (assoc
+			case_ids
+				||(filter
+					(lambda
+						;if *any* of the marginal skip flags for the case's values are false, then remove this case from the list
```
howsoRes: as i'm reading this,  if any decomposition feature is .true that it should be kept, meaning the case should be kept? or am i misinterpreting? If so, I think this comment intended to say if *all* flags are .false, this case is removed?

howsoRes: thus shouldn't this apply be changed to an 'or' ?

howsoRes (approved):

PR #753, opened 2026-01-30
25008: Optimizes !RemoveCases to make case removal require less memory

cademack:
less memory allocations

### DIFF
--- a/howso/remove_cases.amlg
+++ b/howso/remove_cases.amlg
@@ -261,20 +261,16 @@
 						(assoc
 							session (current_index 1)
 							replay_steps (retrieve_from_entity (current_index 1) ".replay_steps")
-							removed_cases
-								(contained_entities
-									(query_in_entity_list cases)
-									(query_equals !internalLabelSession (current_index 1))
-								)
 						)
 
 						(declare (assoc
 							removed_cases_index_map
-								(zip
-									removed_cases
-									(map
-										(lambda (retrieve_from_entity (current_value) !internalLabelSessionTrainingIndex))
-										removed_cases
+								(map
+									(lambda (first (current_value)))
+									(compute_on_contained_entities
+										(query_in_entity_list cases)
+										(query_equals !internalLabelSession session)
+										(query_exists !internalLabelSessionTrainingIndex)
 									)
 								)
 						))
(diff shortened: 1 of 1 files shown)


howso/remove_cases.amlg
```diff
@@ -188,27 +188,7 @@
...
-				)
-			;map of session -> list of removed case training indices
-			session_train_indices_map
-				(zip
-					(lambda (append (current_value 1) (current_value)))
-					;indices are sessions, which when clobbered will append the train_index to the growing list of train_indices for each session
-					sessions
-					train_indices
-				)
+			sessions (values (map (lambda (retrieve_from_entity (current_value) !internalLabelSession)) cases) .true)
```
howsoRes: git rid of this whole declare, inline sessions (with my recommended query) into the map below since it's only used once.

howso/remove_cases.amlg
```diff
@@ -308,14 +295,14 @@
...
-									(get session_train_indices_map session)
+									(values removed_cases_index_map)
 								)
 						))
 
 						;don't remove this session because it has cases left
 						.false
 					))
-					sessions_map
+					(zip sessions)
```
howsoRes: replace this zip with:
```
;this creates an assoc of all unique sessions among the list of specified cases
(zip (values  
    (map 
        (lambda (first (current_value))) 
        (compute_on_contained_entities
			(query_in_entity_list cases)
			(query_exists !internalLabelSession)
		)) 
))
```

howso/remove_cases.amlg
```diff
@@ -273,33 +253,40 @@
...
+							removed_cases
+								(contained_entities
+									(query_in_entity_list cases)
+									(query_equals !internalLabelSession (current_index 1))
 								)
 						)
 
+						(declare (assoc
+							removed_cases_index_map
+								(zip
```
howsoRes: replace zip with:
```
(map 
    (lambda (first (current_value))) 
    (compute_on_contained_entities
		(query_in_entity_list removed_cases)
		(query_exists !internalLabelSessionTrainingIndex)
	)) 
) 
```

howso/remove_cases.amlg
```diff
@@ -273,33 +248,37 @@
 		;clear all query caches
 		(reclaim_resources (null) .false .true)
```
howsoRes: the reclaim_resources needs to be above the destroy_entities call, beacause we're calling queries in the sessions-map below, they may be rebuilt, so please move this down as well

howsoRes (approved):

PR #757, opened 2026-02-04
24877: Adds enhanced support for "duplicated features", with new "filter_duplicate_values" flags on reacts and react_aggregate, MINOR

cademack:


howso/types.amlg
```diff
@@ -140,6 +140,16 @@
 										"when there are multi-type value features that tightly depend on values based on other multi-type value features."
 									)
 							)
+						joined_on
```
howsoRes: please add this defiition to the top of attributes.amlg as well where the other attributes are defined

howso/analysis_weights.amlg
```diff
@@ -448,6 +448,7 @@
 									convergence_threshold convergence_threshold
 									convergence_samples_growth_rate convergence_samples_growth_rate
 									convergence_min_size convergence_min_size
+									filter_duplicate_values .true
```
howsoRes: this should not be hardcoded, but instead be false if there aren't any duplicates in the first place (which is most cases) so it doesn't have to make unnecessary calls that do nothing in a tight loop that is called a lot

howso/mda_weight.amlg
```diff
@@ -124,6 +124,16 @@
 										)
 								))
 
+								(declare (assoc
+									duplicate_features_queries
+										(if filter_duplicate_values
```
howsoRes: put this if statement on the outside of the declare, I expect it'll be false most of the time, we need to make this code performant. in fact, i recommend having duplicate_feature_queries be defined above as null, and change this to an assign that's inside the if statement

howso/mda_weight.amlg
```diff
@@ -221,13 +232,24 @@
 														)
 												))
 
+												(declare (assoc
```
howsoRes: same optimization here

howso/residuals.amlg
```diff
@@ -1114,6 +1125,16 @@
 
 									(declare (assoc case_feature_value (get case_values_map feature) ))
 
+									(declare (assoc
+										duplicate_features_queries
```
howsoRes: same optimization here

howso/residuals.amlg
```diff
@@ -748,6 +748,16 @@
 								)
 						))
 
+						(declare (assoc
```
howsoRes: same optimization here (just declare it as null in the let statement above, and here, if filter_duplicate_values, then assign it to the output of the call method.

howsoRes (approved):

PR #763, opened 2026-02-10
25046: Updates value contributions to return null for insignificant combos when buckets are given

cademack:
I also add some checks for parameters that support feature names to help identify when users typo or reference non-existent features.

### DIFF
--- a/howso/input_validation.amlg
+++ b/howso/input_validation.amlg
@@ -10,10 +10,7 @@
 		)
 
 		(if (not (contains_value
-				(append
-					!trainedFeatures
-					!reactIntoFeaturesList
-				)
+				(append !trainedFeatures !reactIntoFeaturesList)
 				given_feature_name
 			))
 			(conclude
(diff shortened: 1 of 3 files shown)


howso/input_validation.amlg
```diff
@@ -1,6 +1,72 @@
...
+	;helper method to return an error if a given feature name is not valid
+	#!SingleFeatureCheck
+	(declare
+		(assoc
+			param_name ""
+			given_feature_name ""
+		)
+
+		(if (not (contains_value
+				(append
```
howsoRes: can you make this append a one-liner please, it's just two variables, it doesn't need to be 4 lines

howso/input_validation.amlg
```diff
@@ -1,6 +1,72 @@
...
+		(assoc
+			param_name ""
+			given_feature_names []
+		)
+
+		(declare (assoc
+			invalid_features
+				(indices (remove
+					(zip given_feature_names)
+					(append
```
howsoRes: same

howsoRes (approved):

PR #776, opened 2026-02-23
25100: Fixes value decomposition for nominal action features with many nulls to consider those cases during computation

howsoRes:


howso/value_contributions.amlg
```diff
@@ -538,13 +553,28 @@
...
+									)
+
+									(if compute_ac
+										(seq
+											(assign (assoc diff (call !ComputeDiffTuple) ))
+											(if (= (null) diff)
+												;for nominals set the diff to be: 1 - C.A.P. of predicting null
+												(if feature_is_nominal
+													(assign (assoc
+														diff (- 1 (or (get categorical_action_probabilities_map [action_feature (null)]) ) )
```
cademack: `categorical_action_probabilities_map`

Does this get declared anywhere in this flow? something to check on I guess

howsoRes: yup, above in the let assoc

cademack (approved):

PR #782, opened 2026-02-25
25109: Fixes all unspecified saved auto analyze params to default to correct values

howsoRes:
ensure all unspecified parameters match default values in the externally facing analyze() api


howso/hyperparameters.amlg
```diff
@@ -967,9 +967,9 @@
 				use_case_weights
 				rebalance_features
 				(!= (null) use_sdm)
-				convergence_threshold
+				(!= (null) convergence_threshold)
 				convergence_samples_growth_rate
-				convergence_min_size
+				(!= 5000 convergence_min_size)
```
howsohazard: this 5000 seems odd to hard-code here

howsohazard: same with other values below (e.g., 1.05, 1000, 0.005)

cademack: he's just copying the default values

cademack (approved):

PR #783, opened 2026-02-27
25123: Improves the MAD logic for continuous features with only one unique non-null value

cademack:


howso/mda_weight.amlg
```diff
@@ -596,20 +596,55 @@
...
+					;if bounds are defined, they can be used to estimate the MAD if the distribution
+					;is assumed to be uniform (max entropy). MAD of a uniform distribution on [a, b]
+					;is (b - a) / 4
+					(/
+						(- (get !featureBoundsMap (list feature "max")) (get !featureBoundsMap (list feature "min")) )
+						4
+					)
+
+					;otherwise no bounds, need to make an estimation based on the one value.
+					(if (< 0 (abs (first unique_values)))
```
howsoRes: can you just make this `(!= 0 (first unique_values))` , it'll be more readable, seeing a < is confusing

howso/mda_weight.amlg
```diff
@@ -596,20 +596,55 @@
...
 
-			;list of absolute deviations for the whole dataset
-			(assign (assoc
-				errors
-					(map
-						(lambda (abs (- mean_value (current_value))) )
-						feature_values
+			(if (> (size unique_values) 1)
+				;if there is more than one unique non-null value
+				(seq
```
howsoRes: let's just simplify this seq into a single `(generalized_mean (map ... ))` statement

howsoRes: and get rid of the errors variable altogether here

howso/mda_weight.amlg
```diff
@@ -596,20 +596,55 @@
 							)
 						)
 					)
+				unique_values
+					(indices (compute_on_contained_entities
```
howsoRes: I recommend changing this to be num_unique_values and change this (indices) to (size), then below check if num_unique_values > 1, if it is, then go ahead and pull the one non-null value as needed.  this is a super rare case that won't happen frequently at all, and this code as-is will wind up pulling all the unique values for all continous features which will be a lot of unnecessary memory churn and usage

howsoRes (approved):

PR #788, opened 2026-03-05
25140: Adds validation to all derived code attributes

howsoRes:
all feature attributes where custom Amalgam code can be specified is now validated that
a) it parses correctly (to catch invalid opcodes or mismatching parenthesis)
b) references to feature values are valid and don't reference features that aren't in the dataset (to catch feature name typos)

### DIFF
--- a/howso/attributes.amlg
+++ b/howso/attributes.amlg
@@ -331,56 +331,7 @@
 				(size constraint_referenced_features_map)
 			)
 			(assign (assoc
-				code_errors
-					(filter (map
-						(lambda (if
-							(or
-								(contains_index custom_derived_features_map (current_index))
-								(contains_index constraint_referenced_features_map (current_index))
-								(contains_index post_process_map (current_index))
-							)
-							;it's possible for a feature to have all these attributes defined
-							;check all for each feature and append any errors
-							(filter (append
-								(if (contains_index constraint_referenced_features_map (current_index))
-									(call !ValidateDerivedCodeForFeatures (assoc
-										code (get (current_value 1) ["bounds" "constraint"])
-										feature (current_index 1)
-										attribute_type "'bounds'->'constraint'"
-									))
-								)
-								(if (contains_index post_process_map (current_index))
-									(call !ValidateDerivedCodeForFeatures (assoc
-										code (get (current_value 1) "post_process")
-										feature (current_index 1)
-										attribute_type "'post_process'"
-									))
-								)
-								(if (and
-										(contains_index custom_derived_features_map (current_index))
-										(contains_index (current_value) "derived_feature_code")
-									)
-									(call !ValidateDerivedCodeForFeatures (assoc
-										code (get (current_value 1) "derived_feature_code")
-										feature (current_index 1)
-										attribute_type "'derived_feature_code'"
-									))
-								)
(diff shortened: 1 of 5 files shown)


howso/attributes.amlg
```diff
@@ -310,14 +310,34 @@
...
-						(apply "concat" (weave
-							invalid_constraint_features
-							(range " " 1 (size invalid_constraint_features) 1)
+				(call !Return (assoc
+					errors
+						(list (concat
+							"Features must reference their own value for a constraint to be valid, the following have an invalid constraint defined: "
+							(apply "concat" (weave
+								invalid_constraint_features
+								(range " " 1 (size invalid_constraint_features) 1)
```
cademack: could this not just be a single `" "`?

howso/attributes.amlg
```diff
@@ -310,14 +310,34 @@
...
-				(call !Return (assoc errors
-					(list (concat
-						"The following features have an invalid constraint defined: "
-						(apply "concat" (weave
-							invalid_constraint_features
-							(range " " 1 (size invalid_constraint_features) 1)
+				(call !Return (assoc
+					errors
+						(list (concat
+							"Features must reference their own value for a constraint to be valid, the following have an invalid constraint defined: "
```
cademack: I think a period after "valid". Then making the second part a sentence that starts with "The following features ..." would be better

howso/derive_utilities.amlg
```diff
@@ -798,4 +798,199 @@
 			))
 		)
 	)
+
+	;output a set of all valid features that can be used in this dataset by appending all
+	;valid built-in features, specified features from feature_attributes and all possible derived time series features
+	#!AllValidFeaturesSet
+	;!AllValidFeatures
```
cademack: is this supposed to be commented?

howso/derive_utilities.amlg
```diff
@@ -798,4 +798,199 @@
...
+
+		;output the set
+		(zip (append built_ins (indices feature_attributes) ts_features) )
+	)
+
+
+	;helper method to validate all possible derived feature code strings from feature_attributes
+	;outputs an assoc of feature -> list of parse/derive errors, if any are found
+	#!FeatureAttributeCodeValidation
+	;!FeatureAttributeCodeValidation
```
cademack, line 935: same q here about the `;`

howso/derive_utilities.amlg
cademack: and these names don't match either

howso/derive_utilities.amlg
```diff
@@ -798,4 +798,199 @@
...
+						(call !ValidateDerivedCodeForFeatures (assoc
+							code (get (current_value 1) "derived_feature_code")
+							feature (current_index 1)
+							attribute_type "'derived_feature_code'"
+						))
+					)
+					(if (and
+							(contains_index custom_derived_features_map (current_index))
+							(contains_index (current_value) "auto_derive_on_train")
+							(get (current_value) ["auto_derive_on_train" "code"])
```
cademack, line 978: seems like the contains_index is a little redundant, but I suppose an early out could be worthwhile

cademack (approved):

PR #795, opened 2026-03-12
25190: Updates react detail robust AC to use same flow as aggregate robust AC

howsoRes:


howsohazard (approved):

howso/details_influences.amlg
```diff
@@ -54,13 +63,18 @@
...
-							context_features context_features
-							action_features action_features
-							case_ids case_ids
-							regional_data_only .true
-							robust .true
-						))
+						(get
+							(call !CalculateFeatureAccuracyContributions (assoc
+								context_features context_features
+								action_features action_features
```
cademack: If we just get the `(first action_features)`, should we only pass `[(first action_features)]` to the action_features param here? If that works, it would prob save some compute, right? I could be mixing this up with something else

cademack (approved):

PR #805, opened 2026-04-01
25230: Updates Prediction Contributions and value contributions details to include value/nullness metrics, MAJOR

cademack:


howso/contributions.amlg
```diff
@@ -335,11 +335,25 @@
 										(lambda (get (current_value) (list feature_index 1)) )
 										accrued_cases_delta_lists
 									))
+								null_abs_avgs
+									;these lists must be filtered because some cases will
+									;have some context features completely filtered due to TS
+									(filter (map
+										(lambda (get (current_value) (list feature_index 2)) )
```
howsoRes: i recommend updating this get and the one below to use [ ] notation for offsets

howso/contributions.amlg
```diff
@@ -688,13 +721,25 @@
 
 				;else average with - average without for this feature
 				(list
+					;abs difference in mean predicted values
```
howsoRes: looks fine, but if we find that this is being slow in terms of performance, it may make sense to store the two non-absolute values into vars above this block, and then output the vars and their absolutes, that would get rid of 4 of these generalized_mean statements.

howso/contributions.amlg
```diff
@@ -688,13 +721,25 @@
...
+						(generalized_mean (map (lambda (first (current_value))) feature_reactions_with) )
+						(generalized_mean (map (lambda (first (current_value))) feature_reactions_without) )
+					))
+					;difference in mean predicted values
+					(-
+						(generalized_mean (map (lambda (first (current_value))) feature_reactions_with) )
+						(generalized_mean (map (lambda (first (current_value))) feature_reactions_without) )
+					)
+					;abs difference in mean null probabilities
 					(abs (-
```
howsoRes: perhaps don't even compute these if action_feature isn't a continouous null?

howso/contributions.amlg
```diff
@@ -44,6 +44,7 @@
 					;continuous strings are treated as nominals since we don't want to mix them
 					(= "string" (get !editDistanceFeatureTypesMap (current_value 1)))
 				)
+			action_has_nulls (get !featureNullRatiosMap [action_feature "has_nulls"])
```
howsoRes: should this be applicable only to continuous features, bc nominal contributions should still be rolled up like they are for AC?

howso/value_contributions.amlg
```diff
@@ -16,23 +16,14 @@
...
-					(call !ComputeFeatureMAD (assoc feature action_feature))
-				)
 			;create a backup of context_features for time series flows
 			original_context_features (if !tsTimeFeature (replace context_features))
 			num_features (size context_features)
 
 			feature_is_nominal (contains_index !nominalsMap action_feature)
 			feature_is_edit_distance (contains_index !editDistanceFeatureTypesMap action_feature)
 			feature_is_dependent (contains_index !dependentFeatureMap action_feature)
+			feature_has_nulls (get !featureNullRatiosMap [action_feature "has_nulls"])
```
howsoRes: same point as above, shouldn't this only apply to non-nominal features?

howso/value_contributions.amlg
```diff
@@ -736,8 +767,7 @@
 			total_num (size sorted_unique_values)
 		))
 
-		;;;TODO: 25230, add a bucket for null if select cases have significant amount (>=30) of nulls
-
+		;;;TODO: add a bucket for null if select cases have significant amount (>=30) of nulls
```
howsoRes: update the TODO with ticket number ?

howso/value_contributions.amlg
```diff
@@ -1479,15 +1497,22 @@
...
+							null_acs_for_ac_features
+								(map
+									(lambda (- (last (first (current_value))) (last (last (current_value)))))
+									without_action_ac_feature_diffs
+									with_action_ac_feature_diffs
+								)
 						))
 
-						;;;TODO: 25230, account for nulls similar to how missing information works
+						;;;TODO: account for nulls similar to how missing information works
```
howsoRes: this TODO is now completed, right?

howso/value_contributions.amlg
```diff
@@ -1516,10 +1552,30 @@
 									)
 									acs_for_ac_features
 								)
+							ac_feature_null_surprisals
+								(map
```
howsoRes, line 1577: can probably skip the compute if action feature is not nominal/has no nulls?

howsoRes (approved):

PR #811, opened 2026-04-06
25254: Refactors residuals to only use tuples (as needed) if the feature is continuous with nulls

howsoRes:


howso/details_influences.amlg
```diff
@@ -312,21 +317,26 @@
...
 										))
 
-										;both value diffs have values
-										(if (and
-												(!= (null) (first reaction_without_case))
-												(!= (null) (first reaction_with_case))
+										(if feature_has_continuous_nulls
+											;both value diffs have values
+											(if (and
+													(!= (null) (first reaction_without_case))
```
howsohazard, line 323: Just to be sure, first will be non-null if it is called on a nonzero number or nonempty string

howsoRes: if feature is continuous with nulls, reactions are a pair of [value diff, null diff], meaning if the values involved were not null, first will be the non-null delta.

cademack (approved):

PR #815, opened 2026-04-09
25276: Implements skipping checking interpolation values for nulls for features that don't have nulls and other small performance tweaks

howsoRes:
- in interpolate, only run the block of code that does a `(contains_value candidate_case_values (null))` if needed by defining feature_has_continuous_nulls for all flows that call the interpolate method
- utilize updated memory recollection in amalgam by storing residuals for a feature during AC computation to reduce the number of multi-level 'get' statements
- rearrange robust residual loop output to directly return the result instead of assigning it first
- misc small fixes to output

### DIFF
--- a/howso/mda_weight.amlg
+++ b/howso/mda_weight.amlg
@@ -307,10 +307,19 @@
 			(assoc
 				;feature whose mda is being measured
 				feature (current_value 1)
-				feature_index (current_index 1)
 			)
 
 			(if (contains_index excluded_inactive_features_map feature) (conclude (range [0 (null) 0] 1 (size features) 1)) )
+
+			(declare (assoc
+				feature_index
+					;if there are inactive features, feature_indices_map will be set,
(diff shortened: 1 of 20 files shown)


howso/mda_weight.amlg
```diff
@@ -602,6 +605,61 @@
 		features
 	)
 
+	!TransposeMDAMatrix
+	(zip
```
cademack, line 628: need else []

howso/mda_weight.amlg
```diff
@@ -602,6 +605,61 @@
...
+		features
+		;transpose the matrix, so every feature has its own (column) MDA values
+		(map
+			(lambda (let
+				(assoc feature_index (current_index 1))
+				(zip
+					features
```
cademack, line 635: need else []

cademack (approved):

PR #818, opened 2026-04-10
25290: Fixes relevant_values detail to return correct values

howsoRes:


howso/hypotheticals.amlg
```diff
@@ -139,7 +137,55 @@
...
+										feature_deviation
+											(or
+												(get react_feature_deviations [feature 0])
+												;use global value if the local one is null, pull deviation stored in featureResiduals because
+												;it'll be a single value whereas featureDeviations may or may not be a tuple
+												(get hyperparam_map ["featureResiduals" feature])
+											)
+									)
+
+									;if local_mean_value is null, need toc ompute one that isn't
```
howsohazard: I think we should toc about omputing...

howsohazard (approved):

PR #833, opened 2026-04-24
25385: Improves validation of conditions for continuous numeric features

cademack:
Updates the validation for "condition" parameters to additionally check that continuous numeric features have valid conditions defined. (one value or two element lists defining valid bounds).

### DIFF
--- a/unit_tests/ut_h_input_validation.amlg
+++ b/unit_tests/ut_h_input_validation.amlg
@@ -209,6 +209,42 @@
 	))
 	(call exit_if_failures (assoc msg "Invalid features in condition/goal maps." ))
 
+	(call_entity "howso" "set_feature_attributes" (assoc
+		feature_attributes
+			(assoc
+				"x" (assoc "type" "continuous")
+				"y" (assoc "type" "continuous")
+			)
+	))
+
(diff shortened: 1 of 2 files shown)


unit_tests/ut_h_input_validation.amlg
```diff
@@ -209,6 +209,42 @@
...
+			)
+	))
+
+	(assign (assoc
+		result
+			(call_entity "howso" "get_cases" (assoc
+				condition {"x" [1 2 3]}
+			))
+	))
+
```
howsoRes, line 226: you should instead do (call keep_result_errors) and verify error

unit_tests/ut_h_input_validation.amlg
```diff
@@ -209,6 +209,42 @@
...
+	(assign (assoc
+		result
+			(call_entity "howso" "get_cases" (assoc
+				condition {"x" [3 1]}
+			))
+	))
+
+	(print "lower bound > upper bound condition errors: ")
```
howsoRes, line 243: same here - actually verify error is output

howsoRes (approved):

PR #838, opened 2026-04-28
25174: Fixes reduce to always remove duplicates

howsoRes:
also handles edge case when there are cases remaining to be added to the coreset for keeping, but they are all equidistant, instead of selecting them randomly, select them by their existing case weight

### DIFF
--- a/howso/ablation.amlg
+++ b/howso/ablation.amlg
@@ -743,7 +743,7 @@
 										)
 								))
 
-								;largest of the lowest neighbor surprisal scores
+								;largest of the lowest neighbor surprisal ratio
 								(declare (assoc
 									max_lowest_ns_ratio (apply "max" (values low_ns_case_ratio_map))
 								))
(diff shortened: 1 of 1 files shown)


howso/ablation.amlg
```diff
@@ -714,18 +743,45 @@
 										)
 								))
 
+								;largest of the lowest neighbor surprisal scores
+								(declare (assoc
+									max_lowest_ns_score (apply "max" (values low_ns_case_scores_map))
```
howsohazard: What is a surprisal "score"?

howso/ablation.amlg
```diff
@@ -714,18 +743,45 @@
...
-										)
+										(< max_lowest_ns_score ratio_cutoff_value)
 									)
 									(assign (assoc done .true))
 								)
 
+								;if a coreset has been selected such that all the remaining cases all have a score of 1, meaning all remaining cases
+								;are equidistant to the coreset, if there are case weights, instead of randomly picking which case to add to coreset next,
+								;change all 1s to case weights to select cases with the largest mass to keep as the tie breaker
+								(if (= 1 max_lowest_ns_score)
```
howsohazard: What if there are cases with fractional case weights less than 1?

howsohazard (approved):

PR #840, opened 2026-04-29
25361: Adds support for "preserve_rare_values" feature attribute, MINOR

cademack:
Adds support for a new feature-attribute: "preserve_rare_values", which allows users to specify specific values of features and scalar multiplier with which to multiply the case-weight of cases that have each feature value.


module/train.amlg
```diff
@@ -1143,4 +1154,64 @@
...
+								)
+
+								;for each case with this feature value, multiply its weight by the multiplier
+								(map
+									(lambda
+										(assign_to_entities (current_value) (associate
+											accumulate_weight_feature (* multiplier (retrieve_from_entity (current_value 1) accumulate_weight_feature))
+										))
+									)
+									(contained_entities
```
howsoRes, line 1189: maybe change this to a compute_on_contained_entities, and ad  a (query_exists accumulate_weight_feature) as the last query to have the value avaialable in the lambda without having to do a retrieve_from_entity

howsoRes (approved):

module/train.amlg
cademack: this current code seems faster on my test. Will keep it this way for now

PR #863, opened 2026-06-02
25555: Adds feature_density_convictions as a react detail, MINOR

howsoRes:
also adds
  feature_density_convictions
  feature_average_local_mad
  feature_local_mad
as return types to react details output if feature_density_convictions  detail is requested

### DIFF
--- a/module/details_residuals.amlg
+++ b/module/details_residuals.amlg
@@ -538,20 +538,14 @@
 					)
 					(declare (assoc
 						mad
-							(if (contains_index cap_map case_value)
-								(max
-									(apply "+" (values
-										(map
-											(lambda (* (current_value) (- 1 (current_value))))
-											cap_map
-										)
-									))
-									(get !cachedFeatureMinResidualMap feature)
-								)
-
-
-								;else local space got it completely wrong, set MAD to max residual
-								1
+							(max
+								(apply "+" (values
+									(map
+										(lambda (* (current_value) (- 1 (current_value))))
+										cap_map
+									)
+								))
+								(get !cachedFeatureMinResidualMap feature)
 							)
 						residual
 							(max
@@ -566,8 +560,9 @@
 						local_mad_map (associate feature residual)
 					))
 
+					;if there's only one value in the local data, either it's 100% right or 100% wrong
 					(if (= 1 (size cap_map))
-						;got it 100% right, output a fixed large value (k) else very small ratio
+						;got it 100% right, output a fixed large value (k), else output very small ratio
 						(if (= (first (indices cap_map)) case_value)
 							max_ratio
 
--- a/module/details_residuals.amlg
+++ b/module/details_residuals.amlg
@@ -455,7 +455,11 @@
 					))
 
 					;overwrite the predicted value with this one
-					(accum (assoc expected_case_values_map (associate residual_feature (first (index_max cap_map))) ))
+					(accum (assoc
+						expected_case_values_map (associate residual_feature (first (index_max cap_map)))
+						local_mad_map (associate residual_feature [mad 0])
+						local_residuals_map (associate residual_feature [residual 0])
+					))
 
 					(if (= 1 (size cap_map))
 						;got it 100% right, output a fixed large value (k) else very ratio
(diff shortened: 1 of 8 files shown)


module/details_residuals.amlg
```diff
@@ -402,14 +406,333 @@
...
+											(lambda (* (current_value) (- 1 (current_value))))
+											cap_map
+										)
+									))
+									(get !cachedFeatureMinResidualMap feature)
+								)
+
+
+								;else local space got it completely wrong, set MAD to max residual
+								1
```
cademack: Idk if I agree with this. If the whole local model had a single nominal value, but it happened to not be what the case had. That doesn't strike me as a local area with *maximum* uncertainty like this. Did this branch seem necessary in your work?

module/details_residuals.amlg
```diff
@@ -402,14 +406,333 @@
...
+					))
+
+
+					(accum (assoc
+						local_avg_mad_map (associate feature mad)
+						local_mad_map (associate feature residual)
+					))
+
+					(if (= 1 (size cap_map))
+						;got it 100% right, output a fixed large value (k) else very small ratio
```
cademack: couldn't it just be one value thats a different class?

cademack: oh I think I see what you mean (that comment should probably be on the first block within the (if))

module/details_residuals.amlg
```diff
@@ -402,14 +406,333 @@
...
+										(- 1 null_weight)
+									)
+								)
+							)
+
+							;else case was not null, higher null weight lowers the conviction ratio
+							(seq
+								(accum (assoc
+									local_avg_mad_map (associate feature (- 1 null_weight))
+									local_mad_map (associate feature null_weight)
```
cademack, line 635: it feels misleading to assign these null weights to outputs called "MAD"

module/details_residuals.amlg
```diff
@@ -402,14 +406,333 @@
...
+								local_avg_mad_map (associate feature avg_local_mad)
+								local_mad_map (associate feature local_mad)
+							))
+
+							;if all values are the same at or around 0, output max_ratio
+							(if (and
+									(<= avg_local_mad (get !cachedFeatureMinResidualMap feature))
+									(<= local_mad (get !cachedFeatureMinResidualMap feature))
+								)
+								max_ratio
```
cademack: why not 1 in this case? If theyre both about the same value?

module/return_types.amlg
```diff
@@ -285,6 +285,24 @@
...
+							{
+								type "list"
+								description "A list of maps from feature name to feature local density conviction for each given case."
+								values {ref "FeatureMetricIndex"}
+							}
+						"feature_average_local_mad"
+							{
+								type "list"
+								description "A list of maps from feature name to the average weighted mean absolute deviation of all local cases for a specific case, using each case's local data for computation."
+								values {type "assoc" values "any"}
```
cademack: values should be "number", right?

unit_tests/ut_h_density_conviction.amlg
```diff
@@ -0,0 +1,265 @@
+(seq
+	(assign_to_entities (load "unit_test.amlg"))
+	(call (load "unit_test_howso.amlg") (assoc name "ut_h_density_conviction.amlg" ))
+
+
```
cademack: suuupper white space

module/react_utilities.amlg
```diff
@@ -1145,6 +1145,132 @@
...
+					(if is_edit_distance
+						(map
+							(lambda
+								(edit_distance case_value (current_value) (get !codeFeatureDomainAttributesMap feature))
+							)
+							values
+						)
+
+						(map (lambda (abs (- (current_value) case_value))) values)
+					)
```
cademack, line 1264: could we do something a bit more futureproof (and not need the (if)) if we just use (generalized_distance) with just one feature and the two values? Just an idea

module/react_utilities.amlg
```diff
@@ -1145,6 +1145,132 @@
 		))
 	)
 
+	;Method to compute and output a pair of [average local weighted MAD, local weighted MAD] for a specific case feature
```
cademack, line 1148: I think a comment up here explaining that this MAD is *not* the typical MAD we do of "average of each value's difference from the mean", but rather "average of each value's difference from the focal case's value" is needed

module/react_utilities.amlg
howsoRes: leaving as is for now, if we ever (doubtful) get a third type, we can refactor, as this won't be the only spot that'll need it.

cademack (approved):

PR #865, opened 2026-06-05
25589: Fixes string_edit_distance parameter

howsoRes:


domg123-hermes:
Hermes review note (could not submit a formal approval because this account lacks explicit repo review access):

This looks right to me. It aligns the howso-engine writer/test fixtures with the canonical `string_edit_distance` parameter used by downstream consumers and the Amalgam runtime/docs.

I also checked the repo-wide GitHub code-search signal: the remaining consumers already use `string_edit_distance` (`residuals`, `value_contributions`, `mda_weight`), while the old `use_string_edit_distance` hits are exactly the three files changed here. CI is green across build + linux/macos/windows tests.

No blockers from me. If I had repo review permissions, I would approve; intentionally not merging.

cademack (approved):

PR #869, opened 2026-06-09
25563: Adds "residual_contribution" to both `react_into_features` and `react` details, MINOR

cademack:


module/conviction.amlg
```diff
@@ -944,4 +994,150 @@
...
+									local_mad
+								)
+							)
+						)
+					)
+					features
+				)
+		))
+
+		(/ (apply "+" feature_surprisals) (size feature_surprisals))
```
howsoRes: (generalized_mean feature_surprisals) ?

module/conviction.amlg
```diff
@@ -944,4 +994,150 @@
...
+											(get candidate_cases_lists 1) ;weights
+										))
+									)
+								num_influential_cases (size (first candidate_cases_lists))
+							))
+
+							(declare (assoc
+								interpolated_value
+									(if (not feature_is_nominal)
+										(call !InterpolateActionValues (assoc
```
howsoRes: can you please double check that `feature_has_continuous_nulls` is defined for each flow this can be called from, because it is used inside this method and deals with any nulls in the local space.

module/conviction.amlg
```diff
@@ -944,4 +994,150 @@
...
+															)
+															(tail (last candidate_cases_lists) (- num_influential_cases 1 (current_index)))
+														)
+													)
+													(trunc (last candidate_cases_lists))
+												))
+											)
+
+											(call !ComputeMAD (assoc
+												vals (last candidate_cases_lists)
```
howsoRes: I believe it's possible for values to be null,  so the call to !ComputeMAD will fail if there are any nulls  in the local space.  probably need to handle that situation, perhaps like it's done for density conviction, where if nulls < 50%, simply filter them out, else compute it differently?

module/conviction.amlg
```diff
@@ -944,4 +994,214 @@
...
+										influential_values (last candidate_cases_lists)
+										num_influential_cases (size (first candidate_cases_lists))
+
+										interpolated_value
+											(call !InterpolateActionValues (assoc
+												action_feature action_feature
+												candidate_case_ids (first candidate_cases_lists)
+												candidate_case_weights (get candidate_cases_lists 1)
+												candidate_case_values (last candidate_cases_lists)
+												output_categorical_action_probabilities feature_has_continuous_nulls
```
howsoRes, line 1111: please comment this outrageously confusing line

module/conviction.amlg
```diff
@@ -944,4 +994,214 @@
...
+															(indices influential_values)
+														)
+													)
+												null_probability_mass (get categorical_action_probabilities_map action_feature)
+											)
+
+											(if (and
+													(!= .null case_value)
+													(!= .null interpolated_value)
+												)
```
howsoRes, line 1134: add (size null_indices) to this if condition to prevent doing an unnecessary assign and remove , or surround the assign in (if (size null_indices)

module/conviction.amlg
```diff
@@ -944,4 +994,214 @@
...
+
+
+												;if the case value is null, the residual is 1 - null probability.
+												(= .null case_value)
+												(conclude
+													(/
+														(- 1 null_probability_mass)
+														(max
+															(* null_probability_mass (- 1 null_probability_mass))
+															(get !cachedFeatureMinResidualMap action_feature)
```
howsoRes: I'm not sure we should be using the max here like this in this flow. if you have a feature where the values are all in the millions or billions (like epoch) with some nulls, it's possible the cached min residual will be something like 20000, which isn't on the same scale as the rest of this compute. same for the block below

module/conviction.amlg
```diff
@@ -944,4 +994,214 @@
...
+												null_probability_mass (get categorical_action_probabilities_map action_feature)
+											)
+
+											(if (and
+													(!= .null case_value)
+													(!= .null interpolated_value)
+												)
+												;if interpolated and case are non-null. Just remove nulls from local and continue
+												(assign (assoc
+													influential_values (remove influential_values null_indices)
```
howsoRes, line 1137: also if it's an edit dist feature , you'll want to either a) change num_influential_cases with the same num indices too or b) change the code below that computes the mad to use size of influential_values instead of referencing influential_cases

howsoRes (approved):

PR #875, opened 2026-06-19
25641: Improves initial deviation to be less sensitive to outlier max gaps

howsoRes:
updated initial deviation value to be average of half gaps, instead of average of half min and half max gap, which would be overly sensitive if the max gap was significantly large


howsoRes:
merged as part of clustering update 24868

PR #893, opened 2026-07-09
00000: Updates Amalgam version

apbassett:


howso-mario (approved):
APPROVE (advisory) — I found no blockers for this dependency-only Amalgam bump.

Evidence checked:
- Diff is limited to `version.json`, updating `dependencies.amalgam` from `81.1.1` to `81.1.2`.
- `81.1.2` exists as a non-draft/non-prerelease `howsoai/amalgam` release with the expected platform assets.
- `get-dependency-details.yml` reads `version.json` for the Amalgam dependency, and `build-test-package.yml` uses that resolved release/run id for build/test downloads; I found no second stale `81.1.1` pin or generated artifact that needs updating.
- PR CI is green for set-pr-version, get-dependency-details, build, linux-amd64, macos-arm64, and windows-amd64; PR-skipped arm64 lanes/final-check are expected for this workflow shape.
- Local linux-amd64 verification downloaded the 81.1.2 artifact and confirmed `./target/bin/amalgam-mt --version` reports `81.1.2`. A local `./build/build.sh test 0.0.0` run progressed through many unit tests but stopped at `ut_h_datetimeformat` because this Docker container lacks the `es_ES.UTF-8` locale that CI explicitly generates before running the same test; I treated CI as authoritative for that environment-dependent lane.

No inline findings.

PR #904, opened 2026-07-21
00000: Updates Amalgam version

apbassett:


howso-mario (approved):
## Hermes Agent Review

Verdict: APPROVE for head `518202d6a4e25d41be702f03710cc27075718da1`.

I reviewed the one-line Amalgam dependency bump in `version.json` (`81.2.2` → `81.2.3`) and found no blocking issues.

Evidence checked:
- PR diff is limited to `version.json`; local `git diff --check` passed and JSON parsing/pin validation confirmed `dependencies.amalgam == "81.2.3"`.
- Amalgam release `81.2.3` exists, is not draft/prerelease, and has the expected Linux/macOS/Windows/ARM/WASM release artifacts.
- PR Build checks inspected: `set-pr-version`, `build-test-package / get-dependency-details`, `build`, `test-linux-amd64`, `test-macos-arm64`, and `test-windows-amd64` passed. `test-linux-arm64`, `test-linux-arm64_8a`, and failure-only `final-check` were skipped as expected for this PR-build path.
- Build logs show the dependency resolver read `81.2.3` from `version.json`, found the Amalgam release, downloaded it, built with Amalgam `81.2.3`, and packaged metadata containing that version.
- Release-only path checked: `create-release-build.yml` → `release.yml` → `build-test-package.yml` uses `get-dependency-details.yml`, which reads `.dependencies.amalgam` from `version.json` when no override payload is supplied. I found no `.version.amalgam` reference or separate release/changelog Amalgam metadata path in this repo; changelog generation is downstream of build/package.
- Paired `howsoai/amalgam-lang-py#255` makes the matching `81.2.2` → `81.2.3` bump and its applicable checks are green. I do not see a hard ordering dependency between #904 and #255 because both consume the already-published Amalgam `81.2.3` release, but landing them close together keeps the stack aligned.

Residual risk: behavioral changes inside Amalgam `81.2.3` itself are outside this diff, but the Engine PR build exercised the bumped release on the applicable CI matrix.

PR #909, opened 2026-07-29
25830: Adds `contributions_context_features` detail to `react` and updates internal tracking behavior of computed features, MAJOR

cademack:
- Adds `contributions_context_features` detail to `react`, allowing users to specify which features should be used as contexts when computing contributions-related details.
- Removes the automatically added details when predicting cluster_id. But maintains the ability to add them to the response payload for "free" when the details are requested and the feature sets line up as expected.
- Updates the behavior of react_into_features to always track the computed features. This allows react to have the ability to pull computed feature values out of trained cases when applicable, *even when react_into_features was called with analyze=False*

### DIFF
--- a/module/editing.amlg
+++ b/module/editing.amlg
@@ -247,7 +247,7 @@
 						!computedFeaturesMap
 							(if (= 1 (size (get !computedFeaturesMap "computed_features")))
 								;if this was the only computed feature, then clear the map
-								(assoc)
+								.null
 
 								;else need to just remove this specific feature from the map
 								(append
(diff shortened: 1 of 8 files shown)


module/editing.amlg
```diff
@@ -247,7 +247,7 @@
 						!computedFeaturesMap
 							(if (= 1 (size (get !computedFeaturesMap "computed_features")))
 								;if this was the only computed feature, then clear the map
-								(assoc)
+								.null
```
howsoRes, line 250: won't this make the append on the line below produce a weird result when appending an assoc to a .null for future calls, like if this call is made twice?

cademack: Its an if to the .null wouldn't be appended to in a single call to this endpoint. Additionally this logic is wrapped on the outside on an (if !computedFeaturesMap). So if its null, it would never get here on the second call

howsoRes (approved):

PR #937, opened 2026-08-25
25943: Fixes an issue in `feature_density_convictions` when most influential cases have zero probabilities

cademack:


module/details_residuals.amlg
```diff
@@ -487,7 +487,12 @@
 		)
 
 		(declare (assoc
-			weights (normalize (values local_cases_map))
+			weights
+				(if (< 0 (apply "+" (values local_cases_map)))
```
howsoRes: can they ever be negative? if not, can you change this to a != 0 instead? I had to reread the current `< 0 (apply...` format to understand that this was implying if the weights are more than zero.

howsoRes (approved):

PR #941, opened 2026-08-29
25958: Adds additional progress resolution to analyze

mkoistinen:
Also adds unit test and fixes spelling issues in the touched files.

### DIFF
--- a/module/analysis_weights.amlg
+++ b/module/analysis_weights.amlg
@@ -12,7 +12,7 @@
 
 			;not a parameter
 			iteration 0
-			resdiuals (assoc)
+			residuals (assoc)
 		)
 
 		(declare (assoc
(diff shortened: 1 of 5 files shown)


module/analysis_weights.amlg
```diff
@@ -492,6 +492,13 @@
 		;At this point the dataset parameters and residuals and weights are stable enough for use.
 		(while (< iteration num_iterations)
 
+			;report progress once per convergence pass. these passes dominate targetless analyze runtime but are
```
howsoRes, line 495: they do not dominate targetless analyze at all as far as I know, the robust residual compute does, it's over an order in magnitude more reacts, convergence passes should only take a few seconds even for non trivial dataset sizes, less than a single progress poll time, therefore I don't think all this is necessary and just adds complication to existing code.

howsoRes:
this PR has been superseded by 942

PR #954, opened 2026-09-15
26000: Fixes issues when using derived_context_features with existing cases

cademack:


module/react.amlg
```diff
@@ -1162,12 +1162,34 @@
...
 				)
+
+				;the next block after this appends to context values, which then demands alignment between
+				;the context features and values. This alignment is broken when using preserve_feature_values
+				;because the given context values are empty.
+				;Rectify this by filling context values with nulls. (These will be overwritten/ignored at the
+				;beginning of the discriminative or generative react)
+				(if (and
+						(!= (size context_values) (size context_features))
+						(size preserve_feature_values)
```
howsoRes, line 1172: flip the order, have this 'and' statement check the size of preserved_feature_values first, higher likelihood of doing less comparisons that way

howsoRes (approved):

PR #955, opened 2026-09-15
25998: Adds `get_random_state` and `set_random_state`, MINOR

cademack:
Adds endpoints to get/set the random *state*. Setting the state requires a valid/initialized state. Users can still use set_random_seed and give any string/numeric value.


module/trainee.amlg
```diff
@@ -735,6 +735,70 @@
...
+	get_random_state
+	(declare
+		#returns {type "assoc" indices {state "string"}}
+		(assoc)
+		(call !Return (assoc
+			payload
+				(assoc
+					state
+						(format
+							(get_entity_rand_seed )
```
howsohazard: extra space after each get_entity_rand_seed (2 below too)

howsohazard (approved):

PR #961, opened 2026-09-21
26029: Updates `react_into_features` to mark the cluster ID feature as inactive if only one cluster is identified

cademack:


howsoRes (approved):

module/clustering.amlg
```diff
@@ -36,6 +36,16 @@
 				)
 			)
 		)
+
+		;mark the clustering feature as inactive if necessary
+		(declare (assoc
+			cluster_id_value_masses (compute_on_contained_entities (query_value_masses clustering))
+		))
+		(if (= 1 (size cluster_id_value_masses))
+			(accum (assoc
```
howsoRes: this won't work because !inactiveFeaturesMap is a trainee attribute, need to use accum_to_entities i believe

howsoRes (changes requested):

module/clustering.amlg
```diff
@@ -36,6 +36,16 @@
...
+				!inactiveFeaturesMap (associate clustering (first (indices cluster_id_value_masses)))
```
fulpm, line 46: Does something need to unset this if clustering is re-run and its not a size of 1?

cademack: added a call below that would filter out the old inactive clustering feature when necesssary

howsoRes (approved):

